PR #1, opened 2016-03-15
Opus 2015 3dms

jhunkeler:
Implement changes from 2015.3


jhunkeler:
@jaytmiller This looks more manageable than what was in http://www.stsci.edu/~jhunk/diffs/crds.diff

jaytmiller:
Hi Joe,

This doesn't look like stuff which should merge to master,  it looks like stuff which should already be on master/trunk for the most part.  I checked one or two and they appeared to be present.

Todd

jhunkeler:
Hi Todd,

I agree, we should probably just close this out. I remember seeing the 
changes get merged in manually shortly after we migrated this to GH.

Joe

jaytmiller:
We agreed to close this.

PR #23, opened 2016-04-26
Initial commit of new rmaps for NIRSPEC IFU wcs transformation

rij:
components

### DIFF
--- a/crds/jwst/specs/nirspec_ifufore.rmap
+++ b/crds/jwst/specs/nirspec_ifufore.rmap
@@ -1,0 +1,20 @@
+header = {
+    'derived_from' : 'Hand made 2016-04-26',
+    'file_ext' : '.asdf',
+    'filekind' : 'ifufore',
+    'filetype' : 'IFUFORE',
+    'instrument' : 'NIRSPEC',
+    'mapping' : 'REFERENCE',
+    'name' : 'jwst_nirspec_ifufore_0000.rmap',
+    'observatory' : 'JWST',
+    'parkey' : (('META.INSTRUMENT.EXP_TYPE',),),
+    'reference_to_dataset' : {
+        'EXP_TYPE' : 'META.INSTRUMENT.EXP_TYPE',
+    },
+    'sha1sum' : 'f98ed4ca1ae7500542f55c4c7043af2a4f208cab',
+    'suffix' : 'ifufore',
+    'text_descr' : 'IFU fore',
+}
+
+selector = Match({
+})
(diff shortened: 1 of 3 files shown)


jaytmiller:
Hi Robert,

I looked over the specs and they looked decent but I noticed one thing:

    META.INSTRUMENT.EXP_TYPE

should probably be

    META.EXPOSURE.TYPE

Dataset models will require that to match.   The matching header coming from the  dataset will read META.EXPOSURE.TYPE.

What do you think?   Is META.INSTRUMENT.EXP_TYPE just a typo?

One other thing:  if we're translating ASDF EXP_TYPE to META.EXPOSURE.TYPE,  the ASDF itself may not have the ideal organization.   I think very possibly the ASDF should have a META -> EXPOSURE -> TYPE hierarchy.   Unless the DM can hide the ASDF EXP_TYPE.

rij:
OK I'll check this out later.  I just copied an existing spec for a
different ref file type and assumed it was correct.

Robert J.

jaytmiller:
OK,  yeah, two problems:

1.  I think META.INSTRUMENT.EXP_TYPE was a typo from your forebear,  
Bernard.   I fixed the rmaps but historically never go back to the specs 
themselves.

2.  The process I gave for "cloning existing specs" is at odds with 
fixes which happen only on the server.

I now think we should fix all of Bernie's specs to prevent an exact 
repeat.   But...  I'm not sure what to do in general,  not all rmaps on 
the server are specs,  not all specs in the specs directory are 
correct...   Since it's not functionally necessary that the specs be 
correct in all respects,  there is little insurance they will be.

One way to pre-check a new rmap is to do this:

     % python -m crds.certify  ./blah.rmap --verbose

and then wade through the keyword checking output and watch for 
unchecked keys.  It may outright fail or may just show up as UNDEFINED.

Another pre-check is to try a reference file insertion like this:

     % python -m crds.refactor insert   ./blah1.rmap 
./new_madeup_blah2.rmap   asdf_references...

and look at the resulting rmap blah2.   I think these days it will crash 
for this.

Please try those and see if they help identify the problem.  If not 
something else is needed I guess.

Todd

hbushouse:
The data model attribute corresponding to the EXP_TYPE keyword is definitely model.meta.exposure.type   (not instrument)

rij:
I ran certify and checksum on these so they should be OK.
I'll start a new branch to clean up the old rmaps

rij:
There are no data models for the asdf reference files, which is presumably why the reference_to_dataset entry is needed.

hbushouse:
Are there any new types of reference files involved in the NIRSpec IFU WCS ref files? If so, we should add R_nnnnnn keywords to the data model core schema that correspond to any new types of ref files.

rij:
yes, these 3 rmaps are for the new reference file types: ifuslicer, ifufore and ifupost.

rij:
I'll add them to the core schema

jaytmiller:
I confirmed the pull request and installed the new specs on the server.

If you have references and put them someplace global I will check out 
the rmap reference insert logic and submit the empty specs.

Todd

nden:
I'll send a link to reference files shortly.

PR #36, opened 2016-07-18
Added rmap for nirspec photom files

stscirij:


jaytmiller:
Hi Robert,

I merged your pull request but recalled one other thing:  when you convert a .spec into a .rmap and add it to the specs directory,  you should also remove the original .spec file because the new rmap *is* a spec file surrogate.  (The formats differ, but the purpose of declaring the type is the same.   For new types,  the preferred form is .rmap because then the empty rmap is in the VCS and ready to go whenever it's time to add it to CRDS OPS... which very often is months after the work is originally requested and performed,  particularly for HST.)

jaytmiller:
I'll just remove the existing .spec and push the results.

PR #117, opened 2016-10-06
Initial commit of rmaps for new nirspec flats: dflat, fflat and sflat

stscirij:
Added 3 new rmaps for nirspec flats: dflat, fflat and sflat

### DIFF
--- a/crds/jwst/specs/nirspec_dflat.rmap
+++ b/crds/jwst/specs/nirspec_dflat.rmap
@@ -1,0 +1,17 @@
+header = {
+    'derived_from' : 'Hand created 2016-10-06',
+    'file_ext' : '.fits',
+    'filekind' : 'DFLAT',
+    'filetype' : 'DETECTOR FLATFIELD',
+    'instrument' : 'NIRSPEC',
+    'mapping' : 'REFERENCE',
+    'name' : 'nirspec_dflat.rmap',
+    'observatory' : 'JWST',
+    'parkey' : (('META.INSTRUMENT.DETECTOR',),),
+    'sha1sum' : '434caa8a956d77ed4564c980e7d17b7e577f5f52',
+    'suffix' : 'dflat',
+    'text_descr' : 'Detector Flat Field',
+}
+
+selector = Match({
+})
--- a/crds/jwst/specs/nirspec_fflat.rmap
+++ b/crds/jwst/specs/nirspec_fflat.rmap
@@ -1,0 +1,17 @@
+header = {
+    'derived_from' : 'Hand created 2016-10-06',
+    'file_ext' : '.fits',
+    'filekind' : 'FFLAT',
+    'filetype' : 'FORE-OPTICS FLATFIELD',
+    'instrument' : 'NIRSPEC',
+    'mapping' : 'REFERENCE',
+    'name' : 'nirspec_fflat.rmap',
+    'observatory' : 'JWST',
+    'parkey' : (('META.INSTRUMENT.DETECTOR', 'META.INSTRUMENT.FILTER', 'META.EXPOSURE.TYPE'),),
+    'sha1sum' : '9ef65dd9fa556ad82bf675934fef411f2c094b8f',
+    'suffix' : 'fflat',
+    'text_descr' : 'Fore-optics Flat Field',
+}
+
+selector = Match({
+})
--- a/crds/jwst/specs/nirspec_sflat.rmap
+++ b/crds/jwst/specs/nirspec_sflat.rmap
@@ -1,0 +1,17 @@
+header = {
+    'derived_from' : 'Hand created 2016-10-06',
+    'file_ext' : '.fits',
+    'filekind' : 'SFLAT',
+    'filetype' : 'SPECTROGRAPH FLATFIELD',
+    'instrument' : 'NIRSPEC',
+    'mapping' : 'REFERENCE',
+    'name' : 'nirspec_sflat.rmap',
+    'observatory' : 'JWST',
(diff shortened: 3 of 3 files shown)


hbushouse:
I assume there will be imap updates in conjunction with this too? At the very least the nirspec imap obviously needs updating, but because these ref types are not used by other instruments, we'll also need to update their imaps to list these 3 new types as N/A.

stscirij:
Yes, there's still more work to be done before crds is ready to accept these new reference file types.

  Robert J.

________________________________

jaytmiller:
One thing to keep in mind is that unless cal code is data driven by CRDS,  we don't necessarily have to add the other types to CRDS as N/A.   Yesterday Robert and I concluded "not" for these.    The determinant is whether or not a Step is always going to ask CRDS for a reference.   If CRDS is asked,  we need to define the types as N/A where appropriate.   If we know CRDS will not be asked,  we don't have to provide answers for those cases.   

The practice of always defining a type was added for the WCS types so that those Steps do not have to know when they are or are not appropriate to run.  Adding types to imaps as N/A does not necessarily have to be a universal practice.   How do you see these new types working out?

hbushouse:
Right now the new flat_field step that gets applied to data from all instruments lists

reference_file_types = ['FLAT', 'FFLAT', 'SFLAT', 'DFLAT']

The step uses the regular 'FLAT' for all non-NIRSpec instruments, and the new 3 for NIRSpec. So, yes, it's going to result in CRDS being queried for all of these types for all exposures. Now that I think about it, that means the NIRSpec imap would also need an entry for FLAT that say N/A. Unless someone can figure out a better way to do this. I don't know how the assign_wcs step pulls it off (i.e. only listing ref types that it actually needs for a given instrument/mode).

jaytmiller:
I added "classes" structure so the rmap uses UseAfter automatically when updating:

and a corresponding additional tuple of parkeys that apply to UseAfter:

e.g.

-    'parkey' : (('META.INSTRUMENT.DETECTOR',),),
+    'parkey' : (('META.INSTRUMENT.DETECTOR',), ('META.OBSERVATION.DATE', 'META.OBSERVATION.TIME')),

jaytmiller:
OK, wait,  I think the issue is the generic prefetch so I'm going to reverse what I said to Robert.   If a Step runs at all,  I think it will try to prefetch every type whether it's needed by the current instrument or not.   So I think we will indeed need to modify imaps for the other instruments to respond with N/A.   And unless something else changes,  this is a universal issue,  every type used in a generic multi-instrument Step needs to follow suit.   So the notion of always defining a real reference,  or defining a type as N/A,  is "less optional" than I thought;  that's a pattern the prefetch requires unless Steps steer around it somehow.

hbushouse:
Correct. If a step module has "reference_file_types" declaration in it, the pre-fetch will try to find matches for all types listed in that statement, regardless of whether or not the remaining parts of the step ever actually do an explicit call to retrieve a particular ref file. The only way around this in the past was to leave out the "reference_file_types" declaration, in which case no prefetch is done at all and we simply find out whether a given ref file type can be found later on when the step actually does a call to "get_reference_file".

stscirij:
Can you put the reference file types declaration in an if block depending on the exposure type?  The declaration isn't in the init, so doesn't it get executed at import time?  In that case how does the module know the model type?

On Oct 8, 2016 8:18 AM, Howard Bushouse <notifications@github.com> wrote:

Correct. If a step module has "reference_file_types" declaration in it, the pre-fetch will try to find matches for all types listed in that statement, regardless of whether or not the remaining parts of the step ever actually do an explicit call to retrieve a particular ref file. The only way around this in the past was to leave out the "reference_file_types" declaration, in which case no prefetch is done at all and we simply find out whether a given ref file type can be found later on when the step actually does a call to "get_reference_file".

hbushouse:
The "reference_file_types" declaration occurs within the class definition of the step object, but before the definition of the process method, so it's in a part of the code that's unexecutable (at least as far as I can tell, but what do I know?) and hence it's not possible to have it occur with some set of if/else blocks that depend on instrument or mode. Besides it's way before the data model has even been loaded, so you don't even know what instrument/mode you're working with yet. The stpipe infrastructure reads the list of ref type definitions when it instantiates each step class and then does calls to CRDS to query for those ref files. It's either that layer within stpipe that loads the data model and sends it as an argument to the CRDS query, or it just sends the file name and something in CRDS loads it into a data model. At which point you finally know what instrument and mode you're dealing with.

stscirij:
So I guess I don't see how stpipe knows what the type of model is.  Wouldn't it be clearer if we put the ref types declaration in an init for the step class, then call super on that class so the init of the base class is executed as well?
Probably more than we want to worry about right now, easier to just adjust the imaps...

On Oct 8, 2016 1:36 PM, Howard Bushouse <notifications@github.com> wrote:

The "reference_file_types" declaration occurs within the class definition of the step object, but before the definition of the process method, so it's in a part of the code that's unexecutable (at least as far as I can tell, but what do I know?) and hence it's not possible to have it occur with some set of if/else blocks that depend on instrument or mode. Besides it's way before the data model has even been loaded, so you don't even know what instrument/mode you're working with yet. The stpipe infrastructure reads the list of ref type definitions when it instantiates each step class and then does calls to CRDS to query for those ref files. It's either that layer within stpipe that loads the data model and sends it as an argument to the CRDS query, or it just sends the file name and something in CRDS loads it into a data model. At which point you finally know what instrument and mode you're dealing with.

philhodge:
Here's an example:

x.py contains:
(jwstdev)bash-4.1$ cat x.py

class dummy(object):

     reference_file_types = ['extract1d']

     def __init__(self):
         print("dummy")

 >>> import x

 >>> print("{}".format(x.dummy.reference_file_types))
['extract1d']

So reference_file_types is accessible as soon as a module is imported.

Phil

PR #281, opened 2017-06-22
Added JWST MIRI versions of PERSAT, TRAPDENSITY, and TRAPPARS omitted…

jaytmiller:
… from

jira-crds-114,115,116 due to no available references.

### DIFF
--- a/crds/jwst/specs/combined_specs.json
+++ b/crds/jwst/specs/combined_specs.json
@@ -924,6 +924,28 @@
             "tpn":"miri_pathloss.tpn",
             "unique_rowkeys":null
         },
+        "persat":{
+            "derived_from":"handmade spec 06-22-2017",
+            "extra_keys":null,
+            "file_ext":".fits",
+            "filekind":"persat",
+            "filetype":"PERSAT",
+            "instrument":"MIRI",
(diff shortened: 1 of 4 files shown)


hbushouse:
I'm not aware that there is a viable persistence correction yet for the MIRI detectors and once there is, there's no guarantee it'll use the same kinds of reference files as the NIR instruments. So while I guess it doesn't hurt anything to define these types for MIRI, for the near future the imap (or rmap?) should be set up to return N/A for these types when a query/prefetch is performed using a MIRI exposure as input.

Unless @philhodge knows otherwise.

philhodge:
I agree.  For the time being, we should get a file name of "N/A" for MIRI detectors.

jemorrison:
The current plan for the MIRI persistence is to use the same form of the 
correction that was developed for the NIR
detectors. We are going to test using the same correction terms that the 
NIR detectors has submitted to CRDS and then
see if tweaking them a bit will work well enough for MIRI. If it does 
not then we might just want persistence to be
a NO OP for MIRI.
But so far no testing of the NIR results on the MIRI detectors has  been 
done.

Jane

philhodge:
I'd be glad to help when you get to the point of trying a test.

PR #320, opened 2017-08-15
Added baseline NIRSPEC WAVECORR typespec.

jaytmiller:


hbushouse:
This defines wavecorr for use by nirspec, but aren't some other additions needed to make this ref type N/A for all the other instruments, as well as the nirspec modes that don't use it?

PR #325, opened 2017-08-17
Improve unit tests

jaytmiller:


jaytmiller:
Fix for unit test function clear_crds_state().
Fix for crds.list --references.
Added canary test for crds.bestrefs regression mode.

PR #350, opened 2017-09-22
Added TPN support for FGS_ID-STACK X-dimension of 3200

jaytmiller:


crds/certify/validator_helpers.py
```diff
@@ -314,6 +314,10 @@ def nir_filter(instrument, reftype, exp_type):
     else:
         return True
     
+def nir_xdim(exp_type):
+    """Handle weird X-dimensions for NIR detectors,  currently FGS_ID-STACK=3200 not 2048."""
+    return 3200 if exp_type.upper() in ["FGS_ID-STACK"] else 2048
```
hbushouse: Sorry Todd, I gave you incorrect image size info a while back. The FGS ID-STACK images are 2048 x 2304 (as opposed to 3200). Those dimensions are in the conventional nx vs ny (columns vs rows) sense, as opposed to the reversed python dimensions, which would be [2304,2048].

PR #392, opened 2017-11-20
Modified checksum handling so general default is checksums off

jaytmiller:
Typo in crds.sync log message.


jaytmiller:
The big change here is how FITS checksums are handled during CRDS file submission processing.  Historically they've been broadly enabled in CRDS but rarely used in references,  which turned into a nasty data dependent performance problem once large JWST deliveries started showing up with embedded checksums.  This change limits astropy checksum verification to the certify FITS check and also fitsverify... which eliminates all the other certify and rmap refactoring and catalog updating or even file naming header reads.

PR #405, opened 2018-01-15
Nirspec subarray

jaytmiller:


nden:
@jaytmiller The subarray names in the CAL code are `SUBS200A1, SUBS200A2, SUBS200B1, SUBS400A1`. `SUB1600A1` does not exist. (Note the additional `S`. The full list of Nirspec subarray names is [here](https://github.com/STScI-JWST/jwst/blob/master/jwst/datamodels/schemas/core.schema.yaml#L614).

PR #443, opened 2018-03-26
Modified certify --run-fitsverify to demote ASDF extension errors

jaytmiller:
fitsverify reports JWST CAL data models ASDF extensions as errors.  Since these were expected,  CRDS originally demoted these to WARNING messages.   Later it became apparent that they should instead be demoted all the way to CRDS INFO messages to avoid forcing tedious reviews of expected warnings.   

In the process,  I also generalized and streamlined the CRDS code which recategorizes fitsverify output.  In the case of checksum warnings,  the warning messages are recategorized as CRDS ERROR messages instead using the same mechanism.

Although these ASDF extension errors are new, they're resulting due to increased use of the data models when creating references,  so they are an expected feature of future files.

### DIFF
--- a/crds/certify/certify.py
+++ b/crds/certify/certify.py
@@ -562,13 +562,10 @@ class FitsCertifier(ReferenceCertifier):
 
 # -------------------------------------------------------------------------------------------------
 
-EXEMPT_ERRORS = [
-     'Unregistered XTENSION value',
-     ]
-
-ELEVATED_WARNINGS = [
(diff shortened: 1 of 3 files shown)


stscicrawford:
For reference, I'm just adding a link that the ASDF in FITS not working with fitsverify is a bug and should be fixed in one form or another: https://github.com/STScI-JWST/jwst/issues/1807

PR #466, opened 2018-06-01
Fix dependencies

jhunkeler:
I've been mapping out the dependencies for the JWST pipeline and noticed installing CRDS manually with `pip install crds` failed out completely if `astropy` was not already installed. The `requires` argument does "require" a package, but `pip` and `setuptools` will make no effort to install the requested package. Changing it from `requires` to `install_requires` fixes this.

I also added `extra_requires` to handle installing the pipeline or pulling down additional dependencies needed for file submission.

```
#e.g.
$ pip install crds[jwst]
$ pip install crds[submission]
```

`test_requires` was added just to be thorough, however it's only useful if you intend to run `python setup.py test` to run a local test suite... Otherwise it does nothing useful.


jhunkeler:
Ref: https://github.com/STScI-JWST/jwst/issues/1508

PR #504, opened 2018-09-27
Fix setup crds dependency

jaytmiller:
Removes the "import crds" from CRDS setup.py by automatically doing redundant version number updates in the updated save_version script.   This avoids doing heavy imports like astropy and numpy during CRDS setup.py install.


jdavies-st:
🎉

PR #573, opened 2019-04-02
Cmdline error parsing

jaytmiller:


jaytmiller:
Downgraded fatal errors in crds submit to log ERROR messages followed by an exception.  This avoids calling sys.exit() for programmatic API use cases.

Misc pylint updates.

PR #582, opened 2019-04-11
CCD-292: Submission class API

sean-lockwood:
Not sure why `documentation/demos/submission_demo/simple_submission.ipynb` is changing in this PR.

Still lacks auto-confirm feature and the monitor handle return.  Testing of Exception handling is needed.


sean-lockwood:
Not sure what's going on with the unit tests.  One of {OSX, Linux} is failing due to a server connection issue with `jwst-crds.stsci.edu` but the other is ok.

The "passing" test is silently failing the new nose unit test for `rc_submit.Submission()` because `mock` isn't in the environment (used to skip querying the CRDS server when testing the Submission class).  We could simply remote `sumbit/unit_test.py` for now, as it's not adding much.

PR #615, opened 2019-05-01
Looking for Travis failures reported in CCD-421

eslavich:


eslavich:
Seems like whatever was ailing the checksum tests has passed...

PR #632, opened 2019-05-14
CCD-449 Create PDF of User's Guide

jaytmiller:
Updated CRDS User's Guide Makefile based on JWST CAL documentation Makefile to add the "latex" and "latexpdf" rules among others.


jaytmiller:
This PR just adds the capability to build the PDF.   It doesn't provide any approach for distribution nor is the PDF really ready for distribution,  there are numerous formatting issues.    It may however be handy as a developer option,  particularly for doing reviews.

PR #678, opened 2019-08-28
Implement streaming from S3

eslavich:
Here's an approach to streaming mappings/references


eslavich:
The bucket is a flat cache now, and I was able to test this with the modified jwst and ran a Detector1Pipeline successfully.  I have one last question -- where I'm checking for "NOT FOUND" and "N/A" in the filename, am I doing that correctly?

jaytmiller:
I think the way you handled NOT FOUND and N/A is great for a baseline.

PR #695, opened 2019-09-26
Add awscli as an aws extra dependency

eslavich:
I didn't realize awscli was a pip package, this will save users of the S3 plugin from having to install it separately.


jaytmiller:
I spotted this too independently and already merged a similar change.   Closing this.

PR #726, opened 2020-02-24
Allow NIRSpec IRS2 subarray xsize to be 3200

eslavich:


eslavich:
I ran the certifier on all the NIRSpec reference files in the current context and confirmed that the existing values of XSIZE, XSTART, YSIZE, and YSTART all pass the validations in this branch.

PR #727, opened 2020-02-29
Add retries to get_uri_content()   (small S3 files)

jaytmiller:
Implemented retries for utils.get_uri_content()
Renamed envs/s3.sh to jwst-crds-aws.sh and specialized to s3://dmd-test-crds


jaytmiller:
get_s3_uri_content () is failing for WFIRST when called after JWST CAL is initialized.    See WDM-766 for the mind bending traceback.   Given the potential for S3 failures, it might be time to add support for retries to get_uri_content() which also supports https:// and (effectively anyway) file://.

CAVEAT:  I think this failure is a consequence of bugs related to logging and an excessive log level which may be triggered by CAL settings in the release with git hash ec9c583e;   boto3 is somehow getting into DEBUG logging as well as stpipe.   Guessing,  this situation is 100% fatal (not intermittent) due to repr()'s or str()'s which are being triggered only when DEBUG is on.   Further,  the code involved is 100% isolated to boto3 in an otherwise correctly configured environment which works so long as CAL never runs:

`import boto3
s3 = boto3.resource("s3")
obj = s3.Object(bucket_name, key)
binary = obj.get()["Body"].read()`

In other news,  we should investigate toning down CAL logging for JupyterHub,  debug logging doesn't seem great for classes,  particularly since log messages may be glutting the notebook.  This may mean that formal builds being released to the JWST DMS aren't suitable for JupyterHub.

jaytmiller:
I struggled with the related debug quite a bit but eventually concluded that the gigantic tracebacks I was seeing were occurring due to non-fatal errors in the logging system.   Possibly these were related to recoverable failures in boto3 which resulted in debug output.   Due to the volume of debug output however,  the logging itself is effectively fatal,  a blocker for class use.

Given that the boto3 call ultimately appears to succeed,  following the traceback,   I'm withdrawing this as not addressing the root problem.

PR #728, opened 2020-03-04
Don't logger.setLevel(DEBUG) by defaut

jaytmiller:
I modified logging so the CRDS central logger only does setLevel(DEBUG)f the CRDS verbosity level indicates CRDS will actually issue DEBUG messages.   This is to preclude any possibility of turning on DEBUG for CRDS  dependencies.

The current default state of DEBUG is not to be confused with actually outputting DEBUG messages since they are also governed by the condition CRDS_VERBOSITY > 0.   This change merely defers setLevel(DEBUG) until the moment CRDS does issue DEBUG messages due to sufficient verbosity being set in CRDS.

I'm submitting this because collectively (CAL + CRDS + astropy + ...) logging has issues and there are complex chicken-and-egg relationships so there is always FUD that something in CRDS logging is wrong causing excess output we see.

### DIFF
--- a/crds/core/log.py
+++ b/crds/core/log.py
@@ -115,7 +115,7 @@ import contextlib
 DEFAULT_VERBOSITY_LEVEL = 50
 
 class CrdsLogger:
-    def __init__(self, name="CRDS", enable_console=True, level=logging.DEBUG, enable_time=True):
+    def __init__(self, name="CRDS", enable_console=True, level=logging.INFO, enable_time=True):
         self.name = name
 
         self.handlers = []  # logging handlers, used e.g. to add console or file output streams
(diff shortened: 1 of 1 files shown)


jaytmiller:
In terms of reviewing,  one thing I'm hoping for is an opinion about whether-or-not this is totally unnecessary.   Basically I'm worried about DEBUG state inheritance from CRDS to JWST CAL or CRDS dependencies like BOTO3,  it's never clear where DEBUG is getting turned on for these other packages.

eslavich:
I don't think it's necessary, I checked the logging configuration before and after Detector1Pipeline does its thing, and these are the only loggers with an explicit level set:

```
<Logger IPKernelApp (WARNING)>
<AstropyLogger astropy (INFO)>
<MultilineLogger CRDS (DEBUG)>
<MultilineLogger stpipe (INFO)>
<MultilineLogger tweakwcs.linalg (DEBUG)>
<MultilineLogger tweakwcs.tpwcs (DEBUG)>
<MultilineLogger tweakwcs.matchutils (DEBUG)>
<MultilineLogger tweakwcs.linearfit (DEBUG)>
<MultilineLogger tweakwcs.wcsimage (DEBUG)>
<MultilineLogger tweakwcs.imalign (DEBUG)>
<RootLogger root (WARNING)>
```

CRDS isn't messing with the root logger, and I don't see a problem with it electing DEBUG for its own log messages.

jaytmiller:
OK,  thanks,  I'll withdraw this one too.

PR #729, opened 2020-03-04
Ccd 605 add APCORR reference file type for imaging

stscieisenhamer:


stscieisenhamer:
@jaytmiller the ci failure: Is that something that should be handled in the near future?

jaytmiller:
Fixing the CI failure is not on my TODO list.   

My take is that something probably changed in CAL so the CRDS unit test is now failing.  

I was thinking there are three paths to fixing it:
1. Verify CAL is doing the right thing,  and if not,  fix CAL.
2. Add the missing ERROR message to the CRDS unit test expected output.
3. Update the reftype in the test file to something valid.  (May require a different file if there are other 
issues.)

eslavich:
@stscieisenhamer the test failure is due to https://github.com/spacetelescope/jwst/pull/4605 -- the cal code no longer opens a file with an unrecognized extension.  I vote we simply drop that test.

crds/jwst/specs/fgs_apcorr.rmap
```diff
@@ -0,0 +1,73 @@
...
+    'derived_from' : 'Hand made 2019-10-22',
+    'file_ext' : '.fits',
+    'filekind' : 'APCORR',
+    'filetype' : 'APCORR',
+    'instrument' : 'FGS',
+    'mapping' : 'REFERENCE',
+    'name' : 'jwst_fgs_apcorr.rmap',
+    'observatory' : 'JWST',
+    'parkey' : (('META.INSTRUMENT.DETECTOR',), ('META.OBSERVATION.DATE', 'META.OBSERVATION.TIME')),
+    'sha1sum' : '88283a02fec824ae480cf0cc3bb41ca26577dcc3',
```
eslavich: I was wondering [how on earth](https://github.com/spacetelescope/crds/blob/6f0a16f2848709532cd6111aef585ff6905b3205/crds/core/rmap.py#L399) you got this to match...

crds/jwst/specs/fgs_apcorr.rmap
```diff
@@ -0,0 +1,73 @@
+header = {
+    'classes' : ('Match', 'UseAfter'),
+    'derived_from' : 'Hand made 2019-10-22',
+    'file_ext' : '.fits',
+    'filekind' : 'APCORR',
+    'filetype' : 'APCORR',
+    'instrument' : 'FGS',
+    'mapping' : 'REFERENCE',
+    'name' : 'jwst_fgs_apcorr.rmap',
```
eslavich: Do you happen to know if there are any implications to this name field not matching the actual filename?  Looking through the old specs, some match and some don't.

crds/jwst/specs/fgs_apcorr.rmap
stscieisenhamer: Do you mean the actual value being in the repo? Was just [following the directions](https://innerspace.stsci.edu/display/CRDS/Adding+New+Reference+Types). The `install` script runs the hash calculator, which I forget the command at the moment. I run the install locally, so things are all set.

crds/jwst/specs/fgs_apcorr.rmap
stscieisenhamer: No, I do not, but I'll make match.

jaytmiller: As CRDS renames files,  it copies "name" to "dervived_from" and sets "name" to match the generated name.  derived_from is used to easily diff an rmap against its parent.

If you are doing a reference file submission  with automatic rmap generation,  in the final rmap added to CRDS it will reflect the baseline rmap for the update which in turn is related to the context selected on the submission form.   

If you are doing a manual rmap update (download-modify-upload-with-auto-renaming) it should reflect the name of the rmap you downloaded and hence the baseline for the change.  Ideally,  let CRDS do all renaming.

There is a boundary issue so the first .rmap effectively has no parent.  (Strictly speaking,  it is the spec file which is not in CRDS content.)  In this case the main consequence is that  diffs against a parent aren't supported.

jaytmiller:
I have mixed feelings about dropping the test.   The test demonstrates that CRDS can correctly identify ASDF files even if they are  not named .asdf  ...  which is a backup for identifying type based on extension.   Given that the datamodels will fail the file,  the utility of the feature is diminished.    But It's still potentially important in the context  of uploaded files which are assigned opaque temporary names by Django during upload,   hence the alternative of adding the ERROR message to the expected output may be better.  (If the ability to determine ASDF from file contents is eventually lost,  the  server may be impacted.   certify is only one example where file identification based on contents  is potentially used.)

crds/jwst/specs/fgs_apcorr.rmap
eslavich: Yea, I just couldn't figure out how it was possible to checksum a file that also contains the checksum string.

stscieisenhamer:
I have filed ccd-652 concerning the test failure. Otherwise, since it is an unrelated error, I'm going ahead with the merge.

PR #732, opened 2020-03-25
Create initial rmap and TPNs for new synphot reference file

eslavich:
This new reference file contains a list of valid observation mode strings that will be used to perform integration tests on the other synphot files.  The reference will be initially populated with the obsmode strings in ReDCaT's test_pysynphot_tmc.py script.


eslavich:
I'm just going to merge this material with #734, so this PR can be closed

PR #734, opened 2020-03-30
Additional synphot validations plus integration tests

eslavich:
This PR adds support for more automated synphot file delivery, including:

- Support for validating individual values of a table column
- Additional certifier checks for tmgtab, tmctab, tmctab, thermal, and thruput files
- New reference file type, obsmodes, that is a list of valid stsynphot observation modes (see #732)
- Tool for automatically generating new lookup tables
- An "integration test" that confirms consistency among the synphot file types (intended to be run by the server)
- Another integration test that confirms that observation modes listed in the new obsmodes file can be instantiated without error using stsynphot
- Rearrange the validators module into a subdirectory so that special purpose synphot validators can live in their own file

With the corresponding server PR, addresses:

https://jira.stsci.edu/browse/CCD-239
https://jira.stsci.edu/browse/CCD-498


eslavich:
>     1. Is the COMPNAME check good?

I believe so -- everywhere but a subset of the component files, the values are lowercase.  It doesn't make sense to redeliver them all just for that, but I'd like to encourage new and updated files to follow the convention.

>     2. Should DATE be checked?  (old news,  can't remember if it was discussed with ReDCaT.  If tons of files are non-compliant,  when they're modded and resubmitted,  this will be a recurring issue.)

DATE is one of the "comment parkeys" that they requested, so I imagine they wanted it filled in for new deliveries.

> Documentation: Reviewing this I'm wondering how a synphot submission to the server should work... which appears to be very automagical but may have some root assumptions which could be added to the file_submissions.rst docs in the user's guide. Maybe describe how it is supposed to work, which sets of files should be submitted together, etc. Maybe give a screenshot of the new results output and make any relevant comments. It may be as simple as "submit component file + magic == done" but some synphot notes explaining that synphot special case would be helpful.

Yea, excellent point, I'll add this.

PR #737, opened 2020-04-16
Add new ATODSAT column to ACS CCDTAB .tpn

eslavich:
Per Meaghan/Tyler, the column is required and should contain positive integers between 60,000 and 70,000.  I tested the certifier on their new files and saw only warnings about the new row format.


eslavich:
I don't know what's going on with the failing synphot test, but that stuff is all going to get rewritten by #734 anyway, so I'm not going to worry about it.

eslavich:
Meaghan's ready to use this, so I'm going to merge now so I can release it to test.  I'll follow up on any comments in an additional PR.

PR #741, opened 2020-04-22
Added --tpns switch to 'crds list'

jaytmiller:
Dumps out the TpnInfo lines associated with the specified files for use during type development.  See CCD-667.   This is for use during type development.  Basically:

crds list --tpns candidate-reference-file


jaytmiller:
I added this feature to "crds list" to potentially help some during type development or debug.  It shows the total list of TpnInfos which will run against a given file.

PR #744, opened 2020-04-29
Add spec for NIRCam FILTEROFFSET

stscieisenhamer:
Resolves [CCD-665](https://jira.stsci.edu/browse/CCD-665)

### DIFF
--- a/crds/jwst/specs/nircam_filteroffset.rmap
+++ b/crds/jwst/specs/nircam_filteroffset.rmap
@@ -1,0 +1,21 @@
+header = {
+    'derived_from' : 'miri_filteroffset 2020-04-28 16:00:00',
+    'file_ext' : '.asdf',
+    'filekind' : 'filteroffset',
+    'filetype' : 'FILTEROFFSET',
+    'instrument' : 'NIRCAM',
+    'mapping' : 'REFERENCE',
+    'name' : 'jwst_nircam_filteroffset_0000.rmap',
+    'observatory' : 'JWST',
+    'parkey' : (('META.INSTRUMENT.CHANNEL', 'META.INSTRUMENT.MODULE'), ('META.OBSERVATION.DATE', 'META.OBSERVATION.TIME')),
+    'reference_to_dataset' : {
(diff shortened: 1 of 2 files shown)


crds/jwst/specs/nircam_filteroffset.rmap
```diff
@@ -0,0 +1,21 @@
+header = {
+    'derived_from' : 'miri_filteroffset 2020-04-28 16:00:00',
+    'file_ext' : '.asdf',
+    'filekind' : 'filteroffset',
+    'filetype' : 'FILTEROFFSET',
+    'instrument' : 'NIRCAM',
+    'mapping' : 'REFERENCE',
+    'name' : 'jwst_nircam_filteroffset_0000.rmap',
+    'observatory' : 'JWST',
+    'parkey' : (('META.INSTRUMENT.CHANNEL', 'META.INSTRUMENT.MODULE'), ('META.OBSERVATION.DATE', 'META.OBSERVATION.TIME')),
```
eslavich, line 10: What are the consequences of grouping these into pairs?

stscieisenhamer: That I actually don't know and higher authority will need to say. When not, there was a class api error. Most of the other jwst maps are this way also. i recall understanding awhile ago but don't remember now.

jaytmiller: The grouping of parameters follows the nesting of different Selector instances,  in this case (and almost always) the nesting is Match -> UseAfter and the parameters for UseAfter match the data,  not the reference file.  (So USEAFTER is compared to OBS-DATE and OBS-TIME,  modulo possibly misspelling those last two keywords.)   Did I understand the question correctly?

stscieisenhamer:
In general, I would agree about "testing". However, I have been putting these up for review specifically for the conversations that have already happened, such as the parameter grouping and the reference_to_dataset. It has helped in getting to know more aspects of the system for me.

jaytmiller:
OK,  that's fine.   I was really sure if this was helpful or not.

PR #745, opened 2020-04-30
Add ABVEGAOFFSET specs

stscieisenhamer:
Resolves [CCD-669](https://jira.stsci.edu/browse/CCD-669)


stscieisenhamer:
I'm going to merge this, but curious if there is a more efficient way. I did not get a chance to look at the "all" specs. Do the all specs just get updated by the more specific specs, as in `dict.update` like?

PR #749, opened 2020-06-22
Recognize GEIS data files as references

eslavich:
Opening this PR to invoke the tests...


eslavich:
This solves the issue reported to the help desk, and doesn't appear to be causing any problems with the tests.

eslavich:
I took a survey of the calls to `is_reference()` and I don't foresee any issues with the change in status of the GEIS data files.

PR #754, opened 2020-09-08
Change merge_overlaps default to true for all observatories

eslavich:
Our users prefer that merge_overlaps defaults to true for JWST as well.  I'll be running the JWST regression tests with this branch to confirm that nothing goes haywire with this change and will comment here with the results.

EDIT: Regression tests passed no problem.  This seems ready to go.


jaytmiller (approved):
LGTM,  that's what I'd have done.

stscieisenhamer (approved):
Initial tests in DMS test runs have indicated no issues. Looks like this is a go.

PR #757, opened 2020-09-23
Support new WFC3 IMPHTTAB columns

eslavich:
Four new columns were added to each binary table in the UVIS reference file (one of the column names varies by table).

### DIFF
--- a/crds/hst/specs/combined_specs.json
+++ b/crds/hst/specs/combined_specs.json
@@ -2670,10 +2670,7 @@
             "text_descr":"Photometry Keyword Table",
             "tpn":"wfc3_imp.tpn",
             "unique_rowkeys":[
-                "OBSMODE",
-                "DATACOL",
-                "PHOTFLAM1",
-                "PHOTFLAM2"
+                "OBSMODE"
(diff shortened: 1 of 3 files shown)


stscieisenhamer (approved):
Ohhh, brutal.

jaytmiller (commented):
I don't fully understand the changes to unique_rowkeys.  1)  I sort of expected the opposite.  2) unique_rowkeys  has a kind of built-in "optional" behavior where if a column is not present, it is just dropped from the "row id tuple".

eslavich:
> I don't fully understand the changes to unique_rowkeys. 1) I sort of expected the opposite. 2) unique_rowkeys has a kind of built-in "optional" behavior where if a column is not present, it is just dropped from the "row id tuple".

I changed that because OBSMODE alone is the unique key to those rows -- they don't want multiple rows to share the same OBSMODE, even if other column values are different.  Does that rhyme with the change I made, or did I misunderstand something?

jaytmiller:
Nope,  that sounds exactly right,   just make sure they really are unique.

PR #758, opened 2020-09-24
Adjust WFC3 IMPHTTAB .tpn

eslavich:
See note in the .tpn, but we're not going to be able to validate the type of the columns containing nested arrays, because `FITS_rec` is not reporting the correct type for those columns.


eslavich:
I'm going to merge this so that the test delivery can get going, but let me know if you have any concerns and I'll follow up with an additional PR.

PR #759, opened 2020-10-09
Update WFC3 DFLTFILE

eslavich:
This removes an old, unused .spec for WFC3's DFLTFILE and replaces it with an .rmap.


stscieisenhamer (approved):

eslavich:
The test failure on macOS was just due to a dropped connection.

PR #767, opened 2020-10-29
Allow astropy.time.Time and datetime.datetime in UseAfterSelector params

eslavich:
Since Roman is using ASDF for reference files, they have the option to use timestamp objects instead of strings for fields like useafter, and there's been some interest in taking advantage of that.

~~This PR updates `UseAfterSelector` to handle `astropy.time.Time` and `datetime.datetime` values.~~

This PR updates the certifier to convert `astropy.time.Time` to the appropriate string format before proceeding.


stscieisenhamer (approved):

jaytmiller (commented):
The code looks really nice but I don't understand where this is coming from.   

There are multiple issues I should mention in case any aren't obvious:

(1) Historically sub-second granularity for UseAfter has not been a strong need.  Or a need at all.   In my experience the differential in "same-mode" periodic files is more on the order of days or weeks,  not milliseconds.

(2) Unless single parameter UseAfter is just a test convenience,  it's important  to note that  rmaps have two perspectives:  reference file and data file.   The parkeys are written from the data file perspective which historically expressed observation datetime in two parts, date and time.  

(3) Right now every parameter value in an rmap,  and every parameter in an incoming matching "header", is either a string or effectively a string.   Diverging from that may have broader consequences.  The {str: str, ...} nature of headers is a root assumption, and may have some relevance in things like JSONRPC calls if e.g. CRDS evolves to a lambda or something.

(4) One of the grosser aspects of HST development was the plethora of different date/time formats used depending on instrument and/or era,  particularly for USEAFTER.  So I'm wondering if we're not re-opening the door to different kinds of behavior depending on how the date time parameters are actually expressed.   With strings,  there is one definition from the header being compared to one definition in the rmap and it's clear what they both are  and how they behave.  I can imagine someone debugging with a string and getting different results than a problem which originally occurred using a datetime or astropy.Time.

nden:
Just to clarify. This came as a request from the Roman INS team. It will be limited to the Roman CRDS processing as Roman is the only mission working in asdf format only. We will limit it only to `astropy.Time` objects and will require isot format in datamodels which can be easily represented as the expected string as needed. 
Unless there is anticipation of major issues under these limitations I think we should give it a try. We have some time to fix issues with Roman.

eslavich:
After discussing more with Todd, I'm going to change this so that the Time objects are converted to strings before they reach the selector code.  I'll comment again once the PR is ready.

eslavich:
Okay, this is ready to go again.  I updated the description above.

jaytmiller (commented):
Not sure if this is being handled elsewhere,  but heavy_client.py has a  check_parameters() routine guarding getreferences() and getrecommendations() which drops parameters which are not simple types.

eslavich:
> Not sure if this is being handled elsewhere, but heavy_client.py has a check_parameters() routine guarding getreferences() and getrecommendations() which drops parameters which are not simple types.

I was thinking of handling that in stpipe, before getreferences/getrecommendations receives the header, but maybe it would be better for CRDS to do the conversion?  It knows what format it wants, after all.

jaytmiller (approved):
As long as converting Time to a string is handled somewhere,  I have no strong opinions on it.

eslavich:
> As long as converting Time to a string is handled somewhere, I have no strong opinions on it.

We may yet decide to implement this in CRDS, but I'll do that in a separate PR if need be.

PR #768, opened 2020-11-10
Add support for Roman FLAT reference type

eslavich:
This adds support for the Roman FLAT reference file type.


eslavich:
The test failures are due to a recent change in jwst, I'm fixing them in another PR: https://github.com/spacetelescope/crds/pull/769

jbdoggett (approved):

PR #770, opened 2020-11-10
Add some basic tests of Roman support

eslavich:
We added some initial Roman rules files in https://github.com/spacetelescope/crds-cache-test/pull/5, so we should now be able to run some basic tests.  I looked through the existing JWST and HST-specific tests and added Roman versions where appropriate (except for the "certify a mapping" and "certify a reference" test cases that @PaulHuwe is working on).


eslavich:
Since adding romancal broke the Python 3.6 Travis job, I also updated .travis.yml to replace 3.6 with 3.8, and fixed the multiprocessing lock tests that have been failing on 3.8.

PaulHuwe (approved):

PR #771, opened 2020-11-12
Tests for WFI calibration mapping file and reference file validation

PaulHuwe:
Addressing tickets CCD-751 (test 663.4) & CCD-753 (Test 663.6). These tests are to validate the following:

WFI Calibration (CCD-751):
- Certify romancal datamodel validation of a good file.
- Certify romancal datamodel validation failure of a bad file.
- Certify crds tpn validation failure of a bad file.

Calibration Mapping (CCD-753):
- Certify crds confirmation of proper context mapping.
- Certify crds validation failure with duplicate case rules.
- Certify crds checksum failure with duplicate case rules.
- Certify crds tpn validation failure with improper metadata value in a sample rmap file.

Misc. notes:
- All added asdf files are less than 2kb in size.
- Ellipsis are used for the following reasons:
  - To suppress enum lists (so that future developers don't need to cascade tpn/schema enum changes through to many tests). Salient test failure type words preserved.
  - To suppress meta data ID numbers in romancal warnings (since these may change as metadata grows, and aren't the actual error).
  - To suppress slight differences in output between running all tests vs. running tests individually.


crds/tests/test_certify.py
```diff
@@ -800,6 +830,15 @@ def certify_rmap_compare():
     >>> test_config.cleanup(old_state)
     """
 
+def certify_roman_rmap_compare():
```
jbdoggett, line 833: Should there be a test where certifying against a context fails?

crds/tests/test_certify.py
```diff
@@ -800,6 +830,15 @@ def certify_rmap_compare():
...
+    """
+    >>> old_state = test_config.setup(url="https://roman-crds-serverless.stsci.edu", observatory="roman", cache=test_config.CRDS_TESTING_CACHE)
+    >>> certify.certify_file("roman_wfi_flat_0004.rmap", context="roman_0004.pmap")
+    CRDS - INFO -  Certifying 'roman_wfi_flat_0004.rmap' as 'MAPPING' relative to context 'roman_0004.pmap'
+    >>> test_config.cleanup(old_state)
+    """
+
+
 def certify_jwst_bad_fits():
```
jbdoggett, line 842: Do we really want certify tests for all missions in the same file?

eslavich:
I restarted the tests.  The initial failures were due to missing schema updates from https://github.com/spacetelescope/romancal/pull/33, which has since been merged.

crds/tests/test_certify.py
```diff
@@ -843,6 +892,25 @@ def checksum_duplicate_rmap_case_error():
...
+    """
+    >>> from crds.refactoring import checksum
+    >>> old_state = test_config.setup(url="https://roman-crds-serverless.stsci.edu", observatory="roman")
+    >>> checksum.add_checksum("data/roman_wfi_flat_0004_duplicate.rmap")
+    CRDS - INFO -  Adding checksum for 'data/roman_wfi_flat_0004_duplicate.rmap'
+    CRDS - ERROR -  Duplicate entry at selector ('WFI01', 'F158') = UseAfter vs. UseAfter
+    >>> test_config.cleanup(old_state)
+    """
+
+def certify_roman_bad_tpn_value_rmap_case_error():
```
jbdoggett, line 905: I think this is meant to test an invalid keyword value in the rmap, yes? The name of the function reads to me like testing bad values in the tpn file. Something like certify_roman_invalid_rmap_tpn like certify_roman_invalid_asdf_tpn above.

crds/tests/test_certify.py
```diff
@@ -757,6 +757,36 @@ def certify_AsdfCertify_valid():
     >>> doctest.ELLIPSIS_MARKER = '...'
     """
 
+def certify_roman_valid_asdf():
+    """
```
eslavich, line 761: It is possible to add a traditional docstring description at the top of these new tests before the code starts (see `test_certify_check_rmap_updates` for an example).  Maybe we should add some prose to each of these explaining what they test, and also mention that they're testing official Roman requirements?  Just so people know that they can't be refactored away even if they appear to be similar to other existing tests.  @nden do you think we should document the specific requirement number here in the test or is that unnecessary detail?

eslavich (commented):
The tests look great, I just had one suggestion about adding further documentation.

crds/tests/test_certify.py
eslavich: I think I was the one who made the suggestion to work in this file, but I see the appeal of isolating the Roman tests to their own module.  Besides helping us stay organized as we add more and more missions in here, it would also let us run everything Roman in one go.

crds/tests/test_certify.py
sosey: you dont need to necessarily document the requirement number, but definitely add some simple prose explaining what is being tested, and what the input and expected output should be :)

crds/tests/test_certify.py
```diff
@@ -757,6 +757,36 @@ def certify_AsdfCertify_valid():
...
+    >>> old_state = test_config.setup(url="https://roman-crds-serverless.stsci.edu", observatory="roman", cache=test_config.CRDS_TESTING_CACHE)
+    >>> certify.certify_file("data/roman_wfi16_f158_flat_small.asdf", observatory="roman")
+    CRDS - INFO -  Certifying 'data/roman_wfi16_f158_flat_small.asdf' as 'ASDF' relative to context None
+    CRDS - INFO -  Checking Roman datamodels.
+    >>> test_config.cleanup(old_state)
+    """
+
+def certify_roman_invalid_asdf_schema():
```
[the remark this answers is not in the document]
PaulHuwe: I will add comments later today.

crds/tests/test_certify.py
PaulHuwe: @eslavich how do you feel about this? I don't have the best feel for regenerating "general case" type tests in CRDS.

crds/tests/test_certify.py
PaulHuwe: I like moving these to their own file as well, but I think it needs to be its own issue / PR, as it requires moving Ed's tests and recreating the class structure, thus taking it beyond the scope of this PR.

crds/tests/test_certify.py
PaulHuwe: I will rename this today.

crds/tests/test_certify.py
PaulHuwe: I will add comments today.

crds/tests/test_certify.py
eslavich: If we don't need it for the requirements, and it doesn't test anything that is specific to the `crds.roman` module, then I'm inclined not to duplicate the general tests.

crds/tests/test_certify.py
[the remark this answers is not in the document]
PaulHuwe: Comments added.

crds/tests/test_certify.py
PaulHuwe: Issue made.

crds/tests/test_certify.py
PaulHuwe: Comments added.

crds/tests/test_certify.py
PaulHuwe: Test renamed.

crds/tests/test_certify.py
PaulHuwe: Skipped as duplicate.

jbdoggett (approved):

eslavich (approved):
🚀

PR #773, opened 2020-11-13
Tests for WFI spectroscopy calibration reference file validation

PaulHuwe:
Addressing ticket CCD-752 (test 663.5). These tests are to:

Certify romancal datamodel validation of a good spectroscopy file.
Certify romancal datamodel validation failure of a bad spectroscopy file.
Certify crds tpn validation failure of a bad spectroscopy file.

Misc. notes:
- These are very similar to the existing roman calibration tests. 
- All added asdf files are less than 2kb in size.
- Ellipsis are used for the following reasons:
  - To suppress enum lists (so that future developers don't need to cascade tpn/schema enum changes through to many tests). Salient test failure type words preserved.
  - To suppress meta data ID numbers in romancal warnings (since these may change as metadata grows, and aren't the actual error).
  - To suppress slight differences in output between running all tests vs. running tests individually.


eslavich (commented):
This looks good, I'll just wait to merge until Travis reports back.

jbdoggett:
Is it the case that the only thing that distinguishes between an image file and a spectroscopic file is the value of the optical element?

jbdoggett (approved):

PaulHuwe:
Right now, you are correct that the only difference in the metadata is the value of instrument.optical_element. (Obviously the data is different.) As metadata develops, more changes may appear.

eslavich (commented):
LGTM!

PR #774, opened 2020-11-18
Ccd 783 Define all currently know jwst step parameter reference types

stscieisenhamer:
Resolves [CCD-783](https://jira.stsci.edu/browse/CCD-783)


eslavich (approved):
Holy toledo!  I didn't examine all of these but the template looks good to me.

install
```diff
@@ -2,20 +2,26 @@
...
-fi
-
 if [ "$1" = "--skip-types" ]; then
     shift;
     dont_skip_types=0
 else
     dont_skip_types=1
 fi
 
+dev_install=0
```
jbdoggett, line 12: It'd be better to handle setting dev_install the same way that dont_skip_types is. Doesn't matter which way, just consistency helps understanding the code. I personally like the way dev_install is done with its compactness.

jbdoggett, line 12: Comments at the top with regard to usage would be helpful describing what skip-types and dev do. And, also especially since their order matters.

install
```diff
@@ -2,20 +2,26 @@
...
+if [ "$1" == "--dev" ]; then
+    shift;
+    dev_install=1
+fi
+
+# The clean script removes temporary and junk files that should not be
+# committed to the source code repo.
+if [ $dev_install = 0 ]; then
```
jbdoggett, line 20: Shouldn't there be 2 equals signs? (same goes for lines 5, 48, 59 & 68)

stscieisenhamer: Thanks to bash and backwards compatibility, [either one is OK](https://tldp.org/LDP/abs/html/comparison-ops.html).

PR #775, opened 2020-11-19
Add step parameter reftype pars-rscdstep

stscieisenhamer:
Update to include the newly renamed RscdStep.


stscieisenhamer:
going to merge this in since it is another step parameters reftype.

PR #777, opened 2020-12-03
Add test_bad_files tests for Roman

jbdoggett:
Added 3 tests that mirror the existing JWST tests. The new tests depend on changes to crds-cache-test (PR #6) which added a bad reference file and rmap.  Until the crds-cache-test is merged, the tests for this PR are expected to fail.


jbdoggett:
The new tests are redundant with the existing JWST tests.  Nothing being tested is actually observatory specific.

PR #778, opened 2020-12-15
Remove row checking requirement for miri apcorr

stscieisenhamer:
MIRI is changing the format of the APCORR reference files that will no longer be using FITS tables. Remove the column verification.


eslavich (approved):
LGTM

PR #782, opened 2020-12-18
Updates to improve roman/locate.py test coverage.

jbdoggett:
crds/roman/locate.py was updated to improve test coverage. Four problems were uncovered:

1. Decompose_newstyle_name function: The regex match was a no-op, and is corrected to use the fullmatch version for all observatories except for JWST. For JWST, some tests fail with extensive errors too much to fix now, so a comment was put in about it, but the fullmatch function was not implemented there.
2. Mapping value check in the properties_inside_mapping function: For JWST and Roman, map.filekind compared against upper case values was erroneous, and replaced with map.mapping compared against lower case values.
3. For the roman locate_file and locate_dir functions, test cases with mode=None mysteriously causes errors in tests in other modules. The mode=None tests cases are disabled with comments.
4. The tests/test_rmap.py tests depends on the set of imap files in the tests/data directory for which one was added for the roman/locate.py tests. Tests/test_rmap.py was updated to account for the new roman test file.


crds/jwst/locate.py
```diff
@@ -181,7 +181,22 @@ def decompose_newstyle_name(filename):
 
     assert instrument in INSTRUMENTS+[""], "Invalid instrument " + repr(instrument)
     assert filekind in FILEKINDS+[""], "Invalid filekind " + repr(filekind)
+
+    # The original regular expresssion "\d*\" matches anything in the below assert statement, so it's basically a no-op.
```
stscieisenhamer, line 185: Suggest creating an issue now about this so it isn't forgotten.

stscieisenhamer (approved):

crds/jwst/locate.py
jbdoggett: CCD-876 created.

PR #785, opened 2020-12-24
CCD-891 Update nirspec apcorr spec to include filter and grating as parkeys

stscieisenhamer:
Resolves [CCD-891](https://jira.stsci.edu/browse/CCD-891)

### DIFF
--- a/crds/jwst/specs/combined_specs.json
+++ b/crds/jwst/specs/combined_specs.json
@@ -4787,7 +4787,7 @@
                 "Match",
                 "UseAfter"
             ],
-            "derived_from":"Hand made 2019-10-22",
+            "derived_from":"CRDS-395 updates 2020-12",
             "extra_keys":null,
             "file_ext":".fits",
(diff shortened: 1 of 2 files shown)


eslavich (approved):
This LGTM.  Parkey can't be changed by simply delivering a new rmap, right?  It has to be updated in the spec?  Is it the only header field like that, or does most of the header need to be updated in the spec?  (maybe this is a research project I need to embark upon)

stscieisenhamer:
I haven't confirmed code-wise, but experience and ReDCaT experience, say that `crds certify` apparently uses only the spec/tpn. At some level, this makes sense, because an rmap would have many rules, possibly with conflicting parkeys, making the certification process at best ambiguous. It could still work, much like bestrefs, where one could still end up with a match, but that match may be unintended for a new reference.

stscieisenhamer:
And what about all the "required" builds that seem to not be happening?  It does not appear that I can force them to try again?

PR #792, opened 2021-01-28
Add schema_uri feature to rmap header

eslavich:
This PR adds an optional schema_uri field to the rmap header.  When present, any ASDF files with that reftype will be validated against the specified schema by the certifier.  I also dropped the datamodels.open check for pars- reftypes, since we're dropping StepParsModel from jwst.

I need to add certifier tests when schema_uri is present, but it'll be easier to do that later once an .rmap with a schema_uri has actually been delivered.

There is a Big Change (tm) in this that I should call out: with this PR, it is no longer possible to certify a file without access to a CRDS server or cache.


eslavich:
I ran the server tests with this branch on dlhstcrds and dljwcrds, all passed.

stscieisenhamer (approved):

PR #799, opened 2021-02-19
CRDS-416 step parameter references for pipelines where steps are skipped

stscieisenhamer:
Resolves [CRDS-416](https://jira.stsci.edu/browse/CRDS-416)


crds/jwst/specs/fgs_pars-detector1pipeline.rmap
```diff
@@ -0,0 +1,24 @@
...
+    'text_descr' : 'Step parameters for Detector1Pipeline',
+}
+
+comment = """
+Step parameter references are JWST step and pipeline configurations.
+This reftype provides the configuration for jwst.pipeline.Detector1Pipeline
+"""
+
+selector = Match({
+    'N/A' : 'N/A',
```
eslavich, line 23: What's the distinction between specs with a `'N/A' : 'N/A',` rule and those with an empty selector?  The default rule is new to me.

stscieisenhamer: Actually nothing and would like an opinion on this. Usually the initial rmaps are these spec rmaps. Because I did want this rule in the initial rmaps, because if there is not match for any pars-xxx reference, it should not fail but return N/A, I placed it in the specs only to not forget it.

So, it is really just a "remember to do this". It appears to have no effect, but not really sure.

eslavich: Hmm... looks like the selector doesn't make it into combined_specs.json.  Since that's the case, I don't think it can have an effect, because the software doesn't load up the individual files.

I'm slightly in favor of the empty selector but only because it matches our existing spec files and might save a bit of "what is this doing?" confusion in the future.  But I won't protest either way.

crds/jwst/specs/fgs_pars-darkpipeline.rmap
```diff
@@ -0,0 +1,23 @@
...
+    'classes' : ('Match', 'UseAfter'),
+    'derived_from' : 'Hand-made, SCSB, CRDS-416, 2020-10',
+    'file_ext' : '.asdf',
+    'filekind' : 'pars-darkpipeline',
+    'filetype' : 'pars-darkpipeline',
+    'instrument' : 'FGS',
+    'mapping' : 'REFERENCE',
+    'name' : '',
+    'observatory' : 'JWST',
+    'parkey' : ((), ('META.OBSERVATION.DATE', 'META.OBSERVATION.TIME')),
```
eslavich, line 11: Do we need the first component of the key if it's empty?  I mean, is it possible to do this:

```
'parkey' : (('META.OBSERVATION.DATE', 'META.OBSERVATION.TIME'),),
```
then:
```
selector = UseAfter({
})
```

crds/jwst/specs/miri_pars-darkpipeline.rmap
```diff
@@ -1,17 +1,23 @@
...
     'suffix' : 'pars-darkpipeline',
     'text_descr' : 'DarkPipeline runtime parameters',
 }
 
+comment = """
+Step parameter references are JWST step and pipeline configurations.
+This reftype provides the configuration for jwst.pipeline.DarkPipeline
+"""
+
 selector = Match({
```
eslavich, line 22: If you do opt to keep the default match rule, does this selector need one?

crds/jwst/specs/niriss_pars-detector1pipeline.rmap
```diff
@@ -0,0 +1,24 @@
...
+    'text_descr' : 'Step parameters for Detector1Pipeline',
+}
+
+comment = """
+Step parameter references are JWST step and pipeline configurations.
+This reftype provides the configuration for jwst.pipeline.Detector1Pipeline
+"""
+
+selector = Match({
+    'n/a' : 'n/a',
```
eslavich, line 23: Should these be capitalized?

crds/jwst/specs/fgs_pars-darkpipeline.rmap
stscieisenhamer: probably not. just copied from some existing one that did the same thing. however, i like it because if a selector becomes necessary, the placeholder is there, for those of us who don't remember ****.

crds/jwst/specs/miri_pars-darkpipeline.rmap
stscieisenhamer: removing the defaults, so non-issue.

crds/jwst/specs/niriss_pars-detector1pipeline.rmap
stscieisenhamer: removing defaults so non-issue.

crds/jwst/specs/miri_pars-tweakregstep.rmap
```diff
@@ -8,10 +8,11 @@ header = {
...
     'observatory' : 'JWST',
     'parkey' : (('META.EXPOSURE.TYPE',), ('META.OBSERVATION.DATE', 'META.OBSERVATION.TIME')),
-    'sha1sum' : '0c33a939fcc322dbe3f25fd68c2ec9ca383b1464',
+    'sha1sum' : '380734da361762b534b4201cc49006f862b25bab',
     'suffix' : 'pars-tweakregstep',
     'text_descr' : 'TweakRegStep runtime parameters',
 }
 
 selector = Match({
+    'N/A' : 'N/A',
```
eslavich, line 17: Missed this one?

stscieisenhamer: two actually....

eslavich (approved):
LGTM!

PR #800, opened 2021-02-20
Fix checksum issues

stscieisenhamer:


eslavich (approved):
I must confess I did not compute them with pencil and paper.  I'll make a note to add a unit test sometime that checks these...

PR #805, opened 2021-03-02
Add CCDAMP to ACS SATUFILE parkey

eslavich:
It has come to light that the ACS SATUFILE will need to match on CCDAMP for one of the detectors.  We're allowing N/A because the file for the WFC detector will not.


eslavich:
I'm going to merge this so that I can unblock file delivery.  I'll follow up with an additional PR if there are any review concerns.

PR #809, opened 2021-03-17
CRDS-400 Add FULLP to valid list of subarray names for NIRCam

stscieisenhamer:


crds/jwst/tpns/nircam_all.tpn
```diff
@@ -53,7 +53,7 @@ META.SUBARRAY.NAME          H   C   O   SUB128,SUB160,SUB160P,SUB320,SUB320A335R
                                         \
                                         MASKA210R,MASKA430R,MASKBLWB,MASKBSWB,\
                                         \
-                                        FULL,ANY,GENERIC,N/A
+                                        FULL,FULLP,ANY,GENERIC,N/A
```
eslavich, line 56: Do you know if the `crds/jwst/subarrays/NIRCam_Subarrays.xml` file is used for anything?  That's missing FULLP, but it may just be an artifact of earlier development.

eslavich (approved):
LGTM

crds/jwst/tpns/nircam_all.tpn
stscieisenhamer: For completeness, from a slack chat: No indication that these XML files are used in any bit of code. A module titled `hack...` appears to use these to construct TPNs.

Since tests pass, these will be ignored until otherwise indicated.

PR #816, opened 2021-04-07
Add support for nircam extract1d

eslavich:
Errors in the tests clued us into the fact that this rmap was delivered before the spec existed.  I filed a separate issue to investigate how that's possible.


eslavich:
@stscieisenhamer I'm going to merge this immediately so that we can get the tests passing on #815 but I'll follow up with another PR if you have any comments.

PR #818, opened 2021-04-12
Add 5 and 6 to the permissible COS LIFE_ADJ values

eslavich:
COS will be delivering reference files with LIFE_ADJ=5 this year, and LIFE_ADJ=6 in 2022.  I don't see the harm in accepting both early.


stscieisenhamer (approved):
Wow, that is all over the place!

PR #820, opened 2021-04-23
Add retries and optional checksum and file size verification to crds_s3_get

eslavich:
Some useful updates to the crds_s3_get script:

- Add retries.  We've been seeing some transient errors from S3 and this will save pipeline operators from having to investigate and rescue jobs.
- Add optional checksum and file size verification

### DIFF
--- a/scripts/crds_s3_get
+++ b/scripts/crds_s3_get
@@ -2,6 +2,12 @@
 """
 S3 download plugin for CRDS.
 
+Exit status:
+0 - success
+10 - failed file size verification
+11 - failed checksum verification
+other codes - aws-cli failure.  See https://docs.aws.amazon.com/cli/latest/topic/return-codes.html
+
 As of 2021-04-23, copies of this script are maintained in the crds
 and caldp repositories.  Please ensure that any bug fixes make it into
 both!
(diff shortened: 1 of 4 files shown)


scripts/crds_s3_get
```diff
@@ -1,3 +1,76 @@
...
+        else:
+            output = "\\n".join(result.stdout.strip().splitlines())
+            print(f"crds_s3_get: Failed to download '{args.source}' with return code {result.returncode}: {output}", file=sys.stderr)
+            returncode = result.returncode
+
+        next_attempt += 1
+        if returncode == 0 or next_attempt > args.max_retries + 1:
+            sys.exit(returncode)
+        else:
+            # Exponential backoff with jitter, see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
```
stscieisenhamer, line 69: `jitter` a technical term! the telescopes have jitter. this can fly!

stscieisenhamer (approved):
Context of usage may make this clear, but maybe some commentary about return codes.

eslavich:
I only just now noticed that the code that calls the plugin has a retry feature, so I'm going to keep the checksum/file size checks and remove the retry logic from this script.

PR #826, opened 2021-05-21
Support Roman in programmatic submission

eslavich:


crds/submit/background.py
```diff
@@ -72,7 +72,7 @@ def background_complete(thread):
     Returns the result of background task.
     """
     if isinstance(thread, threading.Thread):
-        while thread.isAlive():
```
eslavich, line 75: The camelCase version of this method seems to have been dropped in Python 3.9, but `is_alive` is available in all the versions that we support.

PR #828, opened 2021-06-10
Validate submitted synphot files with both pysynphot and stsynphot

eslavich:
We've seen at least one delivery that passed the tests with stsynphot but caused a problem with pysynphot.  Testing with both will give us confidence that new synphot files will work with both packages.


stscieisenhamer (approved):

eslavich:
Test failures are unrelated and due to recently merged changes in the roman-datamodels package

PR #831, opened 2021-07-09
CCD-1011 Update CRDSCFG for NRS_LAMP handling

stscieisenhamer:
Resolves [CCD-1011](https://jira.stsci.edu/browse/CCD-1011)

The CRDSCFG file has been updated to include a number of special exposure types and consider them "normal" processing.

The file itself is part of a hard-to-maintain system and should be deprecated in the very near future. This PR is considered a temporary fix for current users to keep a base level of operation.

### DIFF
--- a/crds/jwst/gen_system_crdscfg.py
+++ b/crds/jwst/gen_system_crdscfg.py
@@ -99,7 +99,11 @@ class CrdsCfgGenerator:
                 else:
                     log.info("Considering", repr(name), "keep")
                     step = p.step_defs[name]  # class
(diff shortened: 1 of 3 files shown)


eslavich (approved):
LGTM.  The CI failures are the usual Roman suspects.

I see lots of references to .cfg files in the YAML but I don't imagine it's worthwhile to try and replace those with pipeline class names if we're just going to do away with the feature anyway.

stscieisenhamer:
ultimate goal is to remove this file completely also. there will be a future conversation about this.

PR #845, opened 2021-09-29
Change JWST validation errors into warnings

eslavich:
This makes the following changes to the recently added JWST validations:

- DATAMODL validations for specific class names have been disabled.  The .tpn language doesn't easily support warnings for these kinds of checks, and we're going to revive this in a few months anyway, so it seems reasonable to just comment them out for now.  I restored the general warning that occurs when the DATAMODL keyword is missing.

- The asdf library dev version check has been changed from an error to a warning

- The get_data_model_flat_dict function has been reverted to its original state.  This means we're back to the status quo of not validating ASDF files and casting arrays to the schema dtype before FITS validation.  Datamodels validation failures will appear as warnings, as before.  Happily this also means we don't need to install a dev branch of jwst.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+11.4.3 (unreleased)
+===================
+
+- Change JWST validation errors into warnings. [#845]
+
 11.4.2 (2021-09-20)
 ===================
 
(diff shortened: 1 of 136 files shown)


stscieisenhamer (approved):
oh, all that tpn changing. uhhhhggg

LGTM

PR #854, opened 2021-11-19
CCD-1083 trim translations to be specific to roman

ddkauffman:


nden (approved):

eslavich (commented):
The changes look good, but the test failure needs to be addressed

PR #860, opened 2022-02-09
CCD-1110: add photom reftype + pytest

ddkauffman:


nden:
This looks good to me. All the failures are hst related, perhaps COS? Is it a matter of adding the new positions - 5 and 6 to test files?

PR #861, opened 2022-02-15
CCD-1109 added pixelarea reftype + pytests

ddkauffman:

PR #862, opened 2022-02-15
Add LITREF to synphot component tpns

eslavich:
LITREF should be an optional string-type keyword but we'll display a warning when it is missing.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,15 +1,32 @@
+11.9.0 (unreleased)
+===================
+
+HST
+---
+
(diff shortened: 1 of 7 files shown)


nden:
Does the change. in context fix most of the errors?
#861 shows 295 errors and this one only 2.

eslavich:
> Does the change. in context fix most of the errors?

It does I think.  The remaining errors here are due to the new synphot warning introduced in this PR.  Hopefully I'll get some green check marks shortly...

PR #869, opened 2022-04-13
CCD-1161 Implement timeout on CRDS Server network requests

stscieisenhamer:
For JSON RPC calls to the server, the default timeout for network operations is None, which means never time out. Though this has not been an issue before, recent events, as documented in CCD-1161, have shown otherwise.

A new environment, CRDS_CLIENT_TIMEOUT_SECONDS, is implemented, with a default value of 60.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,11 @@ Roman
 
 - added: distortion rmap + tpn [#867]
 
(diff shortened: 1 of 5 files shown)


alphasentaurii (approved):
LGTM

PR #882, opened 2022-05-25
ccd-1142/reason-for-delivery-api-validation

alphasentaurii:
- Minor bugfix checks for invalid (special) chars in "reason for delivery" text submitted via programmatic api. Previous behavior caused the api to "freeze" or timeout with no explanation of what caused the error. The new behavior imitates the web api by returning an error with a descriptive help message if the user tries to submit description text containing anything besides alphanumeric, dashes, underscores, commas or periods.
- Added test for crds.rc_submit.Submission().validate() when 'description' field contains illegal characters

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,11 @@
+11.15.1 (unreleased)
+====================
+
+General
+-------
+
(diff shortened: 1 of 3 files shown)


stscirij (approved):
LGTM

PR #883, opened 2022-05-25
CCD-1186 Update and add specs for all instruments for reftype pars-rampfitstep

stscieisenhamer:
Resolves [CCD-1186](https://jira.stsci.edu/browse/CCD-1186)


alphasentaurii:
Think you need just need to update some tests for pars-rampfitstep otherwise LGTM

alphasentaurii (changes requested):
Just need to add the new filekind to doctest crds.tests.test_reftypes.reftypes_jwst_get_filekinds so that tests pass then this should be good to go.

alphasentaurii (approved):
LGTM

PR #885, opened 2022-06-09
roman/optical-element-translation

alphasentaurii:
- added ref-rmap header translation for p_optical_element
- added getreferences test using new distortion reference type


ddkauffman (commented):
lgtm

nden (approved):

PR #888, opened 2022-06-24
fix/roman-useafter-format-revision

alphasentaurii:
CCD-1195 Useafter string reformats with space instead of T between date and time


stscieisenhamer (approved):
LGTM

PR #891, opened 2022-07-06
Enable bestrefs I/O under Windows

stsci-hack:
These simple changes allows the CRDS operations for **'crds.bestrefs.assign_bestrefs()'**, and any other CRDS function that uses these functions, to operate under Windows while keeping the exact same functionality under all other OS's.

The changes were tested under a version of CRDS which was installed simply using **'pip install -e .'** from a local clone of the CRDS code into an existing Python 3.8 conda environment under Windows 10.  In fact, the revised code was tested by updating a WFPC2 observation using the command: 

**crds.bestrefs.assign_bestrefs(['u5b8060pr_c0m.fits'], sync_references=True)**

The updated file was then able to be processed using STWCS later confirming that everything worked as expected for any CRDS installation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,15 @@ JWST
 
 -  update niriss pars-jumpstep parkeys [#890]
 
+11.16.6 (TBD)
+=============
+
+General
+-------
+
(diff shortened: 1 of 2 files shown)


alphasentaurii (approved):
LGTM

PR #892, opened 2022-07-27
CCD-677/equal-weight-special-case

alphasentaurii:
Resolves CCD-677 where some extended batch file submissions allowed more than one reference file to satisfy a given matching rule, potentially causing pipeline crashes. To prevent users from missing or ignoring this, the warning has been changed to an error (submission will not be completed).


crds/core/selectors.py
```diff
@@ -2035,16 +2035,17 @@ def _validate_raw_key(self, key, valid_values_map):
...
-                warn("-"*40 + "\nMatch case\n",
-                     log.PP(self.match_item(key)),
-                     "\nis an equal weight special case of\n",
-                     log.PP(self.match_item(other)), """
-For some parameter sets, CRDS interprets both matches as equally good.
-See the file submission section of the CRDS server user's guide here:
-    https://jwst-crds.stsci.edu/static/users_guide/index.html
-for more explanation.
-""", "-"*40)
+                # if self._merge_overlaps:
```
stscieisenhamer, line 2038: Unless there is some critical future need, which should be explained, there is no need to keep commented code.

stscieisenhamer (approved):
Except for the removal of commented code, LGTM

crds/core/selectors.py
alphasentaurii: the future need was me remembering to ask about the additional condition `if self._merge_overlaps`... will remove comments and resubmit

PR #898, opened 2022-09-14
CCD-1211/hst-special-case-revert-to-warning

alphasentaurii:
Reversion: "equal weight special case" generates a warning instead of error for HST only


stscieisenhamer (approved):
This is a part of the architecture I have never understood. However, would this been possible to put under the mission-specific code? Forget that terminology.

If that is not a thing, then LGTM.

PR #900, opened 2022-09-20
move build configuration into `pyproject.toml`

zacharyburnett:
`setuptools` now supports the `[project]` table, which is defined by [PEP621](https://peps.python.org/pep-0621).

Additionally, `setuptools` now supports its own entry in `pyproject.toml` called `[tool.setuptools]` (https://github.com/pypa/setuptools/issues/1688, https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration); however, it comes with the following disclaimer:
> Support for declaring configurations not standardized by [PEP 621](https://peps.python.org/pep-0621/) (i.e. the `[tool.setuptools]` table), is still in **beta** stage and might change in future releases.

Support for the `tool.setuptools` table will likely be in beta until `toml` is natively supported with the release of Python 3.11.

Given this, we can attempt to consolidate the build configuration into a single `pyproject.toml` file that can possibly be read by other build systems in the future.


zacharyburnett:
There are a few errors with the configuration before this is ready for review.

alphasentaurii (approved):
Looks good to me, thanks Zach. Let me know when this is ready for final review.

zacharyburnett:
> Looks good to me, thanks Zach. Let me know when this is ready for final review.

It looks like this is ready, though I am not sure if the scripts will be available at runtime; they seem to be on my machine. If we have issues with them, then the solution would be to use PEP621's `project.scripts`, which requires an entry point,  instead of `setup(scripts=)` which uses a shebang'ed file.

alphasentaurii:
> > Looks good to me, thanks Zach. Let me know when this is ready for final review.
> 
> It looks like this is ready, though I am not sure if the scripts will be available at runtime; they seem to be on my machine. If we have issues with them, then the solution would be to use PEP621's `project.scripts`, which requires an entry point, instead of `setup(scripts=)` which uses a shebang'ed file.

Ok cool - I can also do a fresh install with this PR to test that locally as an extra precaution

PR #904, opened 2022-10-18
CCD-1214 roman-pipeline-autoconfirmation

alphasentaurii:
Roman pipeline authorized user can pass the --autoconfirm argument to automatically confirm reference file submissions as long as there are no errors. If there are errors, the submission is automatically canceled.

Minor updates to the file_map attribute for submission api script (no longer need to run "update_file_map()" - the script will try to get the renamed mappings and references even if the submission has not yet been confirmed (i.e. pmaps and imaps not actually generated yet).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+11.16.15 (unreleased)
+=====================
+
+Roman
+-----
+- Automatic confirmation for roman pipeline reference file submissions [#904]
+
 11.16.14 (2022-09-22)
 =====================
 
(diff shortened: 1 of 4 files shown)


stscieisenhamer (approved):
The changelog and pr title imply this is roman only, but it is not clear that is the case?

Also, there was a request for making this restricted in other ways? Is there any of that. IMHO this needs no restrictions. However, if there are, what are they?

alphasentaurii:
> The changelog and pr title imply this is roman only, but it is not clear that is the case?
> 
> Also, there was a request for making this restricted in other ways? Is there any of that. IMHO this needs no restrictions. However, if there are, what are they?

Restricted auth is handled on the server side, I created a PR for that as well. So, it is Roman-only in the sense that currently the only user with sufficient permissions for the --autoconfirm option is the roman pipeline user. (not including superusers)

PR #927, opened 2023-03-15
cicd/python-310-311

alphasentaurii:
- Added python 3.10, 3.11
- Removed python 3.8


alphasentaurii:
Closing this PR - Given that Nose is incompatible with python>=3.10 and no longer maintained, recommended solution is a migration to pytest. Once that is complete we can revisit updating CICD to include testing on py3.10+

PR #928, opened 2023-03-17
CCD-1246 Stale archive report core code

stscieisenhamer:
New module `stale.py` was created to support a request to determine how up-to-date JWST archive data is with respect to CRDS contexts.

The code itself does all the heavy lifting necessary, creating the statistics to make a report. How it was used is demonstrated in a [jupyter notebook](https://gist.github.com/stscieisenhamer/575bc31f331cd7cc9cbb6b3799c0027c).

Operations has requested that a command-line tool be built around this core code. Another issue, linked from CCD-1246, will detail those requirements. This PR is to get the core code into the repo for this further development.

There may well be other ways of determining archive "staleness", which, once in repo, can be discussed.


nden:
How difficult is it to make this mission independent or at least work for Roman as well?

stscieisenhamer:
Should not be too difficult. This would be work done in CCD-1257.

stscieisenhamer:
Since this is not part of the operational code, will go ahead and merge to prepare for further work.

PR #933, opened 2023-05-02
CCD-1267 Copied CRDS structure to test folder.

hbrown-ST:
Ru had showed me this directory that wasn't really being used for anything at this point in time, but he asked me to match the directory structure of CRDS into it. I decided to copy files too, though I don't know how necessary that might be, and I'm sure there is a possibility of any links in the files being broken or pointing to the actual crds files, which is probably ok. 

This ticket would close CCD-1266.


hbrown-ST:
Depending on which ticket is merged first, I expect to have a merge conflict in the CHANGES.rst file, as both PRs are trying to modify the same area. Should be a simple fix though.

CHANGES.rst
```diff
@@ -1,3 +1,12 @@
+11.17.01 (2023-05-02)
```
stscieisenhamer, line 1: As per the other PR, leave the date as "unreleased". The release process will fill in dates automatically.

stscieisenhamer (changes requested):
Without knowing the full plan on migrating to `pytest`, I am a bit leery of the file copy aspect. Though technically this should be OK, the "short neck hairs" are not happy.

Maybe a bit more discussion of how the migration will occur is in order.

CHANGES.rst
hbrown-ST: Ah thanks for that clarification, I'll use that in the other PR.

hbrown-ST:
I'm going to use this PR to instead work on converting bestrefs.py to pytest, as test_bestrefs.py, and place it into crds/test/test_bestrefs/test_bestrefs.py. I'm going to revert the commit I made for copying over files, as that is an eventual goal as part of the epic CCD-1264.

PR #934, opened 2023-05-02
CCD-1266 Removed python 3.8 from ci.yml tests

hbrown-ST:
Python 3.8 tests are guaranteed to fail for any pull requests we make, so I removed it from the ci.yml file. 

This merge will close CCD-1266.


hbrown-ST:
Depending on which ticket is merged first, I expect to have a merge conflict in the CHANGES.rst file, as both PRs are trying to modify the same area. Should be a simple fix though.

CHANGES.rst
```diff
@@ -1,3 +1,14 @@
+11.17.02 (2023-05-02)
+===================
```
nden, line 2: Two conventions about the `Changes` file
- We start with `<version_number> (unreleased)` and add the date when a release is made. At that time it's possible to change the version number depending on what goes in the release, though CRDS versions are done differently @stscieisenhamer and @alphasentaurii should comment on this
- the `===` should be the same length as the title, otherwise it renders incorrectly.

nden (approved):
Other than the comments about the change log looks good.

CHANGES.rst
stscieisenhamer: CRDS "long ago" did major versions for build releases. However, between three missions now, and the aforementioned not exactly matching "semantic versioning", we have been trying to move to that. This should be a developer topic at some point. But, one is usually safe with just bumping the patch version.

stscieisenhamer (approved):
pending fix of the changes format, LGTM

PR #935, opened 2023-05-12
CCD-1267: Begin creating pytests for bestrefs.py 

hbrown-ST:
These changes will close https://jira.stsci.edu/browse/CCD-1267


hbrown-ST:
Obviously more to come but I was just moving slowly with adding tests since I've finally got them working but they feel held together by band aids and Elmer's glue.

test/bestrefs/test_bestrefs.py
```diff
@@ -56,18 +117,18 @@ def test_normalize_id(line, expected):
...
-#def test_only_ids():
-#    """Test should demonstrate normalization of ID list."""
-#    test_brs = br.BestrefsScript()
-#    print(test_brs.args)
-#    print(test_brs.only_ids)
-#    assert test_brs.only_ids == expected
+def test_only_ids():
+    """Test should demonstrate only_ids is set to None."""
+    test_brs = br.BestrefsScript()
+    assert test_brs.only_ids == None
```
zacharyburnett, line 123: are you meaning to use an equality check here?
```suggestion
    assert test_brs.only_ids is None
```

hbrown-ST: Yes thank you.

zacharyburnett:
it looks like the `macos-latest` job was time-capped after 6 hours because it took too long to sync CRDS objects: https://github.com/spacetelescope/crds/actions/runs/5193724665/jobs/9364629513#step:5:4608

alphasentaurii:
> it looks like the `macos-latest` job was time-capped after 6 hours because it took too long to sync CRDS objects: https://github.com/spacetelescope/crds/actions/runs/5193724665/jobs/9364629513#step:5:4608

That can sometimes happen and usually just need to try re-running the job. Although I've never seen it try for 6 hours... in any case can probably hold off on re-running since this is still a draft PR.

test/bestrefs/test_bestrefs.py
```diff
@@ -0,0 +1,556 @@
...
+                             ('12/21/1999 05:42:35', '1999-12-21 05:42:35'),
+                             ('1999-12-21T05:42:35', '1999-12-21 05:42:35'),
+                             ('12-21-1999 05:42', '1999-12-21 05:42:00'),
+                             ('19970114:053714', '1997-01-14 05:37:14'),
+                         ])
+def test_reformat_date_or_auto(line, expected):
+    """Test should return date if argument is none, 'auto' if arg is a variation of 'auto',
+    or a reformatted date otherwise."""
+    assert br.reformat_date_or_auto(line) == expected
+    log.reset()
```
braingram, line 33: Does every test ends with `log.reset()`?

This might be a good opportunity for an [auto-used fixture](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request)

Adding something like the following should make every pytest function in this file automatically reset the log after running.
```python
@pytest.fixture(autouse=True)
def reset_log():
    yield
    log.reset()
```

If this doesn't apply to all functions, removing `autouse=True` and adding it as a function argument should work.

hbrown-ST: Thanks, for the most part it would apply to all tests, and occasionally one or more times mid test. I'll try it out.

crds/bestrefs/bestrefs.py
```diff
@@ -12,6 +12,7 @@
 # ===================================================================
 
 import crds
+import logging
```
stscieisenhamer, line 15: CRDS has a specific log module. Is there a reason for not using that?

hbrown-ST: No not necessarily. This import isn't actually used at all and I think was just an artifact of some testing attempts. I'll comment out unused imports in the next commit.

PR #938, opened 2023-06-14
switch default jwst model to non-deprecated JwstDataModel

braingram:
jwst.datamodels.Datamodel is deprecated and will soon be removed

While working to clean up the DataModel links between jwst and stdatamodels I opened a PR to remove deprecated models from stdatamodels (which jwst imports and exposes at `jwst.datamodels`). The model `jwst.datamodels.DataModel` is deprecated and was slated for removal in stdatamodels 1.5. CRDS appears to use this as some sort of default model (the details are beyond my understanding, a crds expert can hopefully chime in and teach me how the "sausage is made").

As `jwst.datamodels.JwstDataModel` is the replacement for `jwst.datamodels.DataModel`, this PR switches the usage to the non-deprecated model.

I attempted to run tests locally but following a `pip install -e .` within the checked out crds repository jwst tests were failing with a seemingly unrelated:
```
...
INTERNALERROR>     url = URL[:-len(URL_SUFFIX)]
INTERNALERROR> NameError: name 'URL' is not defined
```
I suspect that I did not correctly install crds locally and am hoping the CI for this PR will test the changes.

### DIFF
--- a/crds/jwst/locate.py
+++ b/crds/jwst/locate.py
@@ -55,7 +55,7 @@ def get_datamodels():
     if MODEL is None:
         with log.error_on_exception(
                 "Failed constructing basic JWST DataModel"):
-            MODEL = datamodels.DataModel()
+            MODEL = datamodels.JwstDataModel()
     return datamodels
 
 # =============================================================================
--- a/crds/jwst/schema.py
+++ b/crds/jwst/schema.py
@@ -106,7 +106,7 @@ def _load_schema(schema_name=None):
     """Return the core data model schema."""
     from . import locate
     datamodels = locate.get_datamodels()
-    model = datamodels.DataModel(schema=schema_name)
+    model = datamodels.JwstDataModel(schema=schema_name)
     return model.schema
 
 def _schema_to_flat(schema):


hbrown-ST (approved):

braingram:
Thanks @hbrown-ST for the review.

I'm not sure what the convention is for merging. I don't seem to see a button so feel free to handle this whatever way you see is best. Let me know what I can do to help.

hbrown-ST:
Merging seems fine, but perhaps @stscieisenhamer or @alphasentaurii would like to weigh in.

stscieisenhamer (approved):
As long as it is a drop-in replacement, which it appears to be, there should be no issue.

braingram:
> As long as it is a drop-in replacement, which it appears to be, there should be no issue.

Thanks! I should have linked this earlier. `JwstDataModel` is the parent class of the previously used `DataModel`. `DataModel` is otherwise an empty class (that raises a deprecation warning).
https://github.com/spacetelescope/stdatamodels/blob/134252b7f5d6ccf9bed65d0318bd6de5a6a66266/src/stdatamodels/jwst/datamodels/model_base.py#L58

alphasentaurii (approved):
LGTM

PR #941, opened 2023-06-23
HST/WFC3: Added new WFC3 SATUFILE reference file type (CRDS-691)

stscirij:
Added new WFC3 SATUFILE reference file type


alphasentaurii (approved):
LGTM

PR #949, opened 2023-08-29
pytest/core-substitutions

alphasentaurii:
- added pytests for crds.core.substitutions
- additional fixtures and markers

### DIFF
--- a/test/core/test_substitutions.py
+++ b/test/core/test_substitutions.py
@@ -1,0 +1,178 @@
+import pytest
+import logging
+
+# ==================================================================================
+
+from crds.core import log
+from crds.core import substitutions
+
+# ==================================================================================
+
+log.THE_LOGGER.logger.propagate=True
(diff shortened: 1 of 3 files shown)


test/core/test_substitutions.py
```diff
@@ -0,0 +1,178 @@
...
+
+from crds.core import log
+from crds.core import substitutions
+
+# ==================================================================================
+
+log.THE_LOGGER.logger.propagate=True
+log.set_verbose(50)
+
+@pytest.mark.substitutions
```
hbrown-ST, line 14: So in order to add the mark "substitutions", you need to add it to pyproject.toml?

hbrown-ST (approved):

test/core/test_substitutions.py
alphasentaurii: Yes, all markers have to be added there

PR #961, opened 2023-09-14
Completed pytests and test conversions tests for tests/test_locking.py

hbrown-ST:
Got the locking tests done. Still having ttat slight logging issue but I think as long as it works and is testing the same functionality it can be figured out at a later date.


test/core/test_locking.py
```diff
@@ -0,0 +1,131 @@
...
+
+"""
+    assert out_to_check in out
+    assert status == "enabled, multiprocessing"
+    default_shared_state.cleanup()
+
+
+@mark.locking
+def test_multiprocessing_locking(default_shared_state, capsys):
+    _ = config.LOCKING_MODE.set("multiprocessing")
```
stscieisenhamer, line 60: If the return is not used in anyway, there is no reason to explicitly save it to nothing. For me, it confuses readability.

stscieisenhamer (approved):
Just a IMHO commentary about unused function returns. But otherwise LGTM

test/core/test_locking.py
hbrown-ST: Likewise, I'll space that return.

PR #966, opened 2023-09-20
Refactor setup_test_cache to allow for simply updating local cache

stscieisenhamer:
The script `setup_test_cache` always clears out the local cache completely. The command line argument is added to the script to allow only updating the local cache. The usage now is as follows:
```
./setup_test_cache CRDS_TEST_ROOT <c/u>
```
where `<c/u>` is an optional argument. If `c`, the default behavior of completely clearing and rebuilding the local cache is done. This is also the default behavior if no argument is given. If anything other that `c` is given, such as `u`, the cache will only be updated with the current state of the HST/JWST operational servers.

When run without any arguments, a help text is given.

### DIFF
--- a/setup_test_cache
+++ b/setup_test_cache
@@ -1,9 +1,9 @@
 #! /bin/bash -x
 usage="Usage: setup_test_cache CRDS_TEST_ROOT <c/u> where 'c' means clean and 'u' means update"
-update=${2:-"n"}
+update=${2:-"c"}
 export CRDS_TEST_ROOT=${1?$usage}
 
-if [ "$update" = "n" ]; then
+if [ "$update" = "c" ]; then
     (cd $CRDS_TEST_ROOT; rm -rf crds-cache-test;  git clone https://github.com/spacetelescope/crds-cache-test.git;)
     (cd $CRDS_TEST_ROOT; rm -rf crds-cache-default-test; mkdir crds-cache-default-test;)
 fi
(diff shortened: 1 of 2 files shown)


setup_test_cache
```diff
@@ -1,9 +1,12 @@
...
+#! /bin/bash -x
+usage="Usage: setup_test_cache CRDS_TEST_ROOT <c/u> where 'c' means clean and 'u' means update"
+update=${2:-"c"}
+export CRDS_TEST_ROOT=${1?$usage}
 
-export CRDS_TEST_ROOT=$1
-
-(cd $CRDS_TEST_ROOT; rm -rf crds-cache-test;  git clone https://github.com/spacetelescope/crds-cache-test.git;)
-(cd $CRDS_TEST_ROOT; rm -rf crds-cache-default-test; mkdir crds-cache-default-test;)
+if [ "$update" = "c" ]; then
```
hbrown-ST, line 6: should be a '==', no?

hbrown-ST: Looks good otherwise.

alphasentaurii (approved):
LGTM

setup_test_cache
stscieisenhamer: Nope, that is the way the `[` operator works. [Reference here](https://linuxize.com/post/how-to-compare-strings-in-bash/#comparison-operators).  Yes, it is bash and opaque is the best description.

hbrown-ST: Ah interesting, I was looking at [this page](https://ryanstutorials.net/bash-scripting-tutorial/bash-if-statements.php) to make sure and saw `
    #!/bin/bash
    # elif statements
    if [ $1 -ge 18 ]
    then
    echo You may go to the party.
    elif [ $2 == 'yes' ]
    then
    echo You may go to the party but be back before midnight.
    else
    echo You may not go to the party.
    fi
`

as an example. If it works, it works.

PR #967, opened 2023-09-20
Fix test_certify errors

stscieisenhamer:


stscieisenhamer:
The issues addressed were resolved in #968

PR #970, opened 2023-09-22
Converted bad_files to pytest.

hbrown-ST:
Got tests working for bad_files, though expecting some failures since I don't think my container setup is identical to Ru's, and we seem to be getting different results when tests are run.

### DIFF
--- a/test/test_bad_files.py
+++ b/test/test_bad_files.py
@@ -1,0 +1,156 @@
+import logging
+import os
+import crds
+from crds.core import log, utils, config, exceptions
+from crds import client
+from crds.bestrefs import BestrefsScript
+from crds.tests import test_config
+from pytest import mark
+
+# ensure CRDS logger propagates events to pytest log capture.
+log.THE_LOGGER.logger.propagate = True
+
+# Contrived headers which will select bad files.
+
+HST_HEADER = {
+    'INSTRUME': 'ACS',
+    'REFTYPE': 'PFLTFILE',
+    'DETECTOR': 'SBC',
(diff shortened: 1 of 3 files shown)


test/test_bad_files.py
```diff
@@ -0,0 +1,164 @@
...
+    with caplog.at_level(logging.INFO, logger="CRDS"):
+        BestrefsScript(args)()
+        out = caplog.text
+        print('out is')
+        print(out)
+    default_shared_state.cleanup()
+    out_to_check = f"""No comparison context or source comparison requested.
+No file header updates requested;  dry run.  Use --update-bestrefs to update FITS headers.
+ ===> Processing /home/runner/work/crds/crds/test/data/hst/j8btxxx_raw_bad.fits
+Failed processing '/home/runner/work/crds/crds/test/data/hst/j8btxxx_raw_bad.fits' : Failed computing bestrefs for \
```
hbrown-ST, line 116: Ended up having to change this path to '/home/runner/work/crds' to work for runtests but in my container (and pycharm) I get 'home/developer'. Is there something I can modify to get what would be expected here when I run my local tests?

stscieisenhamer: Note that the actual check, below is `<something> in <something>`, which means that you can modify the relevant lines in the expected output to have only a subset of that line.

I see in the original doctest-based that there were ways of flagging ignore areas. This level of complexity is not necessary.

stscieisenhamer (approved):
besides the machine-specific issue, LGTM.

PR #971, opened 2023-09-25
Convert `diff` to pytest

zacharyburnett:
@hbrown-ST how do I get the requisite data files (are they in `master` somewhere?)


alphasentaurii:
> @hbrown-ST how do I get the requisite data files (are they in `master` somewhere?)

All the data is inside test/data/{observatory}, and each has its own fixture. Most of the test data referenced in this module is for HST, so you would just use the `hst_data` fixture, and pass that in as the start of any paths referencing it

zacharyburnett:
four `diff` tests left failing:
```
test/test_diff.py ..F....F.FF..
```

hbrown-ST (approved):

PR #973, opened 2023-09-26
use GitHub Actions cache to cache test cache

zacharyburnett:
It appears the that Pytest CI downloads the same test cache every time; perhaps this can be sped up by caching it per-context


alphasentaurii (approved):
This is great. If you could go ahead and rebase from pytest-migration to get those two failing tests to pass, LGTM

PR #978, opened 2023-09-29
pytest/bestrefs-test-class

alphasentaurii:
- revised pytest ini to look for "Test*" class tests instead of "*Test"
- converted Bestrefs test class and methods to pytest
- added new hst_persistent_state
- renamed crds/tobs/tests classes to "Tests_" so they're not picked up by pytest
- yield state cfg fixtures and perform cleanup automatically regardless of test outcome
- remove cleanup() step from all tests
- added 'mode' kwarg as Config option
- explicitly set mode to 'local' in handful of tests that fail when mode='remote'


stscieisenhamer (approved):
Question: If the `crds/tobs/tests.py` are no longer running, should this module just be removed?

alphasentaurii:
> Question: If the `crds/tobs/tests.py` are no longer running, should this module just be removed?

Probably, unless they are another set of tests that should be migrated to pytest. I'm assuming they were part of the original nose test suite but were not known of because of their location inside the tobs submodule

hbrown-ST (approved):

PR #980, opened 2023-09-29
Converted tests for test_sync.py

hbrown-ST:
Many of the tests for test_sync.py converted rather easily, but I do notice that some tests are not running exactly the same way each time. I imagine it has to do with the set up I'm using, so hopefully that can be solved.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -150,6 +150,18 @@ def hst_shared_cache_state(crds_shared_group_cache):
     cfg.cleanup()
 
 
+@fixture()
+def hst_temp_cache_state(test_temp_dir):
+    cfg = ConfigState(
+        cache=str(test_temp_dir),
+        url="https://hst-crds.stsci.edu",
+        observatory="hst",
+    )
+    cfg.config_setup()
+    yield cfg
+    cfg.cleanup()
+
+
 @fixture(scope='function')
 def jwst_no_cache_state():
     #os.environ["CRDS_MAPPATH_SINGLE"] = test_data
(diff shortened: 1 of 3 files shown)


test/bestrefs/test_bestrefs.py
hbrown-ST, line 1: As Ru had suggested, I commented out the class here since it was likely to fail based on the modification I made to pyproject.toml so that classes for pytest would work.

alphasentaurii: Yes - the replacement for it is in #978

stscieisenhamer (commented):
Should the test_bestrefs changes be removed? They are not properly part of the PR right?

alphasentaurii (approved):
Tentatively approving with a caveat that the configuration for this class will need to be revised pending removal of CRDSTestConfig from conftest which uses a Nose-based setup/teardown and is essentially a temporary artifact brought over from the old test config. Frankly I'm surprised this is working at all but since it is running the tests and they are passing, I'm ok with merging this for now.

alphasentaurii:
> Should the test_bestrefs changes be removed? They are not properly part of the PR right?

These should be removed followed by a rebase. For an example on how to implement a test class, this is now working (and has been merged into pytest-migration. Note that it is not necessary to use CRDSTestCase (which has been deleted):
https://github.com/spacetelescope/crds/blob/212036868a450e3e9a1dd3c211d4079a235eb161/test/bestrefs/test_bestrefs.py#L450

PR #983, opened 2023-10-03
Got tests working for newcontext

hbrown-ST:
Got the new context tests converted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,36 @@ on:
     - pytest-migration #temp
 
 jobs:
+  data:
+    name: retrieve current CRDS context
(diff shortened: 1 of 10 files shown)


alphasentaurii (approved):
Not a blocker for merging but just FYI standard pytest practice is to also add a marker for the module this is in, i.e. refactoring. See other test mods inside refactoring as an example (we can add this in later it's not a big deal)

hbrown-ST:
> Not a blocker for merging but just FYI standard pytest practice is to also add a marker for the module this is in, i.e. refactoring. See other test mods inside refactoring as an example (we can add this in later it's not a big deal)

Alright I added those.

PR #985, opened 2023-10-04
various minor improvements to CI

zacharyburnett:
~runs tests in parallel with `pytest-xdist`,~ add `workflow_dispatch` to `ci.yml` for manually starting tests, and also a `concurrency` block to manage multiple pushes to the same branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,10 @@ on:
     # - master
     - pytest-migration #temp
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   data:
     name: retrieve current CRDS context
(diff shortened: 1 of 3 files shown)


stscieisenhamer (commented):
are the failing tests related to this change?

zacharyburnett:
> are the failing tests related to this change?

I don't think so; the tests run, but some fail intermittently

stscieisenhamer:
Looking again. The two errors, both for test_check_archive_file_api_*, indicate that the argument `--cov=crds` is being passed to the crds call `CheckArchiveScript` as the `prog` argument to `ArgumentParser`. This is indeed not correct, but have no clue as to how that is occurring.

zacharyburnett:
> Looking again. The two errors, both for test_check_archive_file_api_*, indicate that the argument `--cov=crds` is being passed to the crds call `CheckArchiveScript` as the `prog` argument to `ArgumentParser`. This is indeed not correct, but have no clue as to how that is occurring.

appears to have been fixed in 4fc7f6fea4b7c50f3cf41ded14a399f8301dd2d8; I'm not sure why though

PR #988, opened 2023-10-05
pytest/synphot

alphasentaurii:
- Converted synphot tests to pytest: test/misc/test_synphot.py
- Converted synphot lookup generator test class to pytest (added to above test module)

Note: Many of the original nose tests were very long, so I broke them up into individual tests where it made sense to do so. 

Resolves CCD-1352 and CCD-1353


stscieisenhamer (approved):
Despite the cache issue in the CI, LGTM

PR #997, opened 2023-10-18
also test with Python `3.12`

zacharyburnett:


braingram:
The CI test failures appear related to asdf 3.0. I'm looking into these now. So far I've found a few issues:
- `asdftool diff` raises an exception when comparing a tree with an array nested within a list. See: https://github.com/asdf-format/asdf/pull/1672
- the `asdftool diff` output has changed in asdf 3.0 (and changes again in the above linked PR). Previously asdf output an unhelpful `ndarrys differ by contents` with no context as to which arrays differed. The crds test suite is specifically checking for this output:
https://github.com/spacetelescope/crds/blob/d5d92a1c937d13f4a80360b9de71935e835e0d40/test/test_diff.py#L181 and will need to be updated for asdf 3.0 (and 3.0.1 or whatever bug fix includes the fix to the asdftoold diff exception issue linked above)
- there are crds test data files that contain invalid block indexes, for asdf < 3 there was no indication that these were invalid, asdf 3.0 introduces a warning, as the crds tests check for the number of warnings reported this is leading to test failures for asdf 3.0

stscieisenhamer (approved):

stscieisenhamer:
is this still relevant?

PR #1002, opened 2023-10-26
Enable python 3.12 testing in the CI, enable dependabot for CI workflow version checking

braingram:
This PR adds python 3.12 tests to the CI.

While updating the workflow I noticed an out-of-date workflow version. I updated the version and added a dependabot configuration file to this PR that should enable dependabot to automatically open PRs that will update the workflow versions when a new one is available.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+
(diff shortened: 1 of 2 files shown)


braingram:
Shoot, this is mostly a duplicate of #997

PR #1004, opened 2023-10-27
Update tests for asdf 3.0

braingram:
This PR addresses a number of crds test and asdf 3.0 incompatibilities. See comments below for details.

PR #1005, opened 2023-10-27
add asdf upper pin until 3.0 issues can be resolved

braingram:
As described in #997 (see: https://github.com/spacetelescope/crds/pull/997#issuecomment-1783221215) there are multiple issues with the crds test suite and asdf 3.0. Additionally asdf 3.0.0 has a bug in the asdftool diff output which appears to be used by crds.

This PR adds an upper pin to asdf 3.0 until these issues can be resolved.

Alternatively, #1004 addresses the asdf 3.0 incompatibilites. It is currently blocked by https://github.com/asdf-format/asdf/pull/1672 and a release of asdf 3.0.1. If other crds changes are needed adding this pin is a faster option to allow the CI to pass. However it may create issues for jwst and romancal as it would prevent them from installing asdf 3.0.


braingram:
Closing in favor of #1004

PR #1006, opened 2023-10-30
Testing dev tests with numpy 2.0

hbrown-ST:


pyproject.toml
```diff
@@ -2,6 +2,7 @@
 name = "crds"
 description = "Calibration Reference Data System,  HST/JWST/Roman reference file management"
 authors = [{ name = "STScI CRDS s/w developers" }]
+requires-python = ">=3.11"
```
WilliamJamieson, line 5:
```suggestion
requires-python = ">=3.9"
```

Since we support 3.9+

hbrown-ST: At the moment I can't get numpy 2.0 to be picked up without being in python 3.11.

WilliamJamieson: These tests are not even picking up `numpy 2.0`. The requirements file included in this PR does not appear to be doing anything.

Note that if you look at the [dev wheels archive](https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/numpy/) you are linking to in it. There very much are wheels for 3.9 and 3.10. Indeed this is true for both numpy and scipy.

hbrown-ST: Right I found that numpy 2.0 wasn't getting picked up after some digging. Looking into that.

PR #1007, opened 2023-10-30
replace ast.Str with ast.Constant

braingram:
`ast.Str` is deprecated in python 3.12. Creating an instance of `ast.Str` in python 3.12 returns an instance of `ast.Constant` and as the docs state:
https://docs.python.org/3/library/ast.html
```
Changed in version 3.8: Class [ast.Constant](https://docs.python.org/3/library/ast.html#ast.Constant) is now used for all constants.
...
Deprecated since version 3.8: Old classes ast.Num, ast.Str, ast.Bytes, ast.NameConstant and ast.Ellipsis are still available, but they will be removed in future Python releases. In the meantime, instantiating them will return an instance of a different class.
```
As crds only appears to support python 3.9+ this PR replaces `ast.Str` with `ast.Constant`.


stscieisenhamer (approved):

braingram:
Thanks for the review. FWIW I do not appear to have merge permissions so feel free to merge this whenever it suits.

braingram:
> This is minor but the next release version should be 11.17.14, not 11.17.4

Thanks for finding my mistake. I rebased this PR and fixed the changelog.

alphasentaurii (approved):

PR #1013, opened 2023-11-08
(CCD 1385) Added PIXAR headers to MIRI PHOTOM

hbrown-ST:
Missing keywords in delivered miri photom reference files did not trigger an error with crds certify. This change should fix that.

### DIFF
--- a/crds/jwst/specs/combined_specs.json
+++ b/crds/jwst/specs/combined_specs.json
@@ -1196,6 +1196,35 @@
             "tpn":"miri_fringefreq.tpn",
             "unique_rowkeys":null
         },
+        "gain":{
+            "derived_from":"cloning tool 0.05b (2013-04-12) used on 2013-10-04",
+            "extra_keys":null,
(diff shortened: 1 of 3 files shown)


crds/jwst/specs/combined_specs.json
alphasentaurii, line 1: Isn't this supposed to be a MIRI PHOTOM tpn/spec modification?

hbrown-ST: Yeah it is, that was added after running install again. Weird.

alphasentaurii (approved):
LGTM pending confirmation that combined_specs.json are acceptable

crds/jwst/specs/combined_specs.json
hbrown-ST: Recreating the spec file didn't change a thing. I think this just didn't get picked up when I updated miri_gain.rmap for some reason in an earlier PR. Going to merge.

PR #1021, opened 2024-01-23
fix jwst schemas link in documentation

braingram:
The link to the jwst datamodels schemas in the documentation is currently broken:
https://hst-crds.stsci.edu/static/users_guide/reference_conventions.html?highlight=jwst#keywords-for-new-file-formats

This PR updates it to the current location of the schemas:
https://github.com/spacetelescope/stdatamodels/tree/main/src/stdatamodels/jwst/datamodels/schemas

### DIFF
--- a/documentation/crds_users_guide/source/reference_conventions.rst
+++ b/documentation/crds_users_guide/source/reference_conventions.rst
@@ -31,7 +31,7 @@ metadata tree.
 
    .. group-tab:: JWST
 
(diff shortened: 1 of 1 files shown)


braingram:
@hbrown-ST @alphasentaurii I don't have permissions to set labels or request reviewers.

I don't think the PR needs a changelog entry but please let me know if you'd like me to add one.

alphasentaurii (approved):
LGTM. I added the label to skip changelog entry check.

PR #1022, opened 2024-01-24
explicitly specify CRDS in `actions/checkout` to enable reuse of caching workflow in other repos

zacharyburnett:
If other repositories (such as `stenv`) wish to run CRDS tests, and want to cache this testing data, then they can reuse the `cache.yml` workflow by calling
```yaml
jobs:
  crds_test_cache:
    uses: spacetelescope/crds/.github/workflows/cache.yml@master
```

However, currently the `actions/checkout` action does not specify which repo to checkout, meaning that when `cache.yml` is called in another repo, it will checkout THAT repo instead of `spacetelescope/crds`. This can be fixed by adding
```yaml
jobs:
    - uses: actions/checkout@v3
      with:
        repository: spacetelescope/crds
```

### DIFF
--- a/.github/workflows/cache.yml
+++ b/.github/workflows/cache.yml
@@ -39,6 +39,8 @@ jobs:
       key: ${{ steps.key.outputs.key }}
     steps:
       - uses: actions/checkout@v3
+        with:
+          repository: spacetelescope/crds
       - id: key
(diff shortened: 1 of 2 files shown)


alphasentaurii (approved):
thanks Zach

hbrown-ST (commented):
LGTM but the changes.rst file just wants the pr number.

PR #1028, opened 2024-02-20
Crds-230 Adding rmap for nirspec_drizpars

hbrown-ST:
Created a new rmap for nirspec drizpars and removed old spec file.


hbrown-ST:
Tests passed on HST and JWST Dev despite the same CI errors.

PR #1032, opened 2024-02-22
MODEL was added to PEDIGREE verification for JWST.

hbrown-ST:
An rmap for NIRISS NRM needs to have MODEL available for META.PEDIGREE. This change should ideally allow that.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,7 +3,7 @@
 
 JWST
 ----
--Added MODEL to be a PEDIGREE option for JWST. [#1031]
+-Added MODEL to be a PEDIGREE option for JWST. [#1032]
 
 11.17.18 (2024-02-21)
 =====================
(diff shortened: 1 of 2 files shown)


hbrown-ST:
Tests on JWST and HST Dev servers passed, merging.

nden:
Does this need to be added to documentation somewhere? IIRC this column is exposed on the server.

hbrown-ST:
> Does this need to be added to documentation somewhere? IIRC this column is exposed on the server.

I'm not sure about that, would be happy to update any docs if that is the case. I'll check with Jonathan. I made this change because there was a JWST file being delivered by REDCAT that had MODEL for PEDIGREE and was giving certification errors, and the errors are not happening anymore.

stscieisenhamer:
Sent this already to Hunter, but posting here for the record:

> All the jwst reference file documentation is in the jwst repo. INS/ReDCaT should be filing issues/updating the docs. Either ask ReDCaT to do the filing or file yourself but assign to them.

PR #1034, opened 2024-04-04
roman/decorate-old-tests

alphasentaurii:
CCD-1438:
- Added fixtures to conftest.py to resolve ongoing test suite failures:
    - default_cache: $CRDS_PATH or /tmp/crds-cache-default-test 
    - `hst_default_cache_state`, `jwst_default_cache_state`: these use the new `default_cache` fixture to point directly to `crds-cache-default-test`, in addition to explicitly setting mode='auto', url, and observatory. Some tests were failing because they relied on the `crds_shared_group_cache` fixture which returns the CRDS built-in `config.get_crds_path()`. The path returned is not always what the test expects it to be and in many cases was incorrect (e.g. `crds-cache-test` or  'data/'. 

Note: ultimately we may want to move remaining tests away from using the 'crds_shared_group_cache' altogether - I elected not to do that for now because the majority of tests using it were fine and I'd rather not break something that didn't need fixing.

- Decorated tests with observatory-based markers according to the test configuration and/or data being used in the test. For tests that don't use an observatory-based test configuration, the marker is "multimission". The primary intention here is to aid in troubleshooting setup/teardown problems. E.g. a test with an "hst" marker might be a mission-agnostic feature of CRDS, but it explicitly relies on the hst server, hst test data, etc.

- added CRDS_PATH and CRDS_TESTING_CACHE to github actions ci.yml env

CCD-1434:
Added metrics_logger decorator to Roman tests delivered with older builds (test_certify.py)
- DMS4: 
    - test_certify_roman_valid_asdf
    - test_certify_roman_invalid_asdf_schema
    - test_certify_roman_invalid_asdf_tpn
- DMS5:
    - test_certify_roman_valid_spec_asdf
    - test_certify_roman_invalid_spec_asdf_schema
    - test_certify_roman_invalid_spec_asdf_tpn
- DMS6:
    - test_certify_roman_rmap_compare
    - test_certify_roman_duplicate_rmap_case_error
    - test_checksum_roman_duplicate_rmap_case_error
    - test_certify_roman_invalid_rmap_tpn


CHANGES.rst
```diff
@@ -1,3 +1,11 @@
+11.17.20 (unreleased)
+=====================
+
+ROMAN
+-----
+- Added metrics_logger decorator to Roman tests delivered with older builds [#1034]
```
zacharyburnett, line 6:
```suggestion
- Added `metrics_logger` decorator to Roman tests delivered with older builds [#1034]
- Marked individual tests with their respective observatory [#1034]
```

zacharyburnett (approved):
LGTM!

PR #1039, opened 2024-05-08
CCD-1257: Formalize the stale-by-context report tool

stscieisenhamer:
Resolves [CCD-1257](https://jira.stsci.edu/browse/CCD-1257)

This PR formalizes the determination of archive staleness through examination of contexts and affected dataset reports.
The form of the command is
```
$ python -m crds.bestrefs.stale -h
```
See the code for further documentation.

Note that during development, the JWST Mission Office has decided to reprocess the JWST archive based on "data releases" and not on every new CRDS context, making this tool defunct. As such this tool will not be used. Adding this to the repo for documentation purposes mostly.


alphasentaurii (approved):
LGTM

PR #1045, opened 2024-06-10
Added pars-darkcurrentstep rmaps for NIRISS, NIRCAM and NIRSPEC.

hbrown-ST:
This PR handles CCD-1467 which added 3 rmaps for the pars-darkcurrentstep reference type.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+11.17.24 (20204-06-10)
+=====================
+
+JWST
+----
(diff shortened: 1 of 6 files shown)


alphasentaurii (commented):
Should the combined_specs.json file be updated as well? Usually this is done by running './install' inside the crds repo top level directory

hbrown-ST:
> Should the combined_specs.json file be updated as well? Usually this is done by running './install' inside the crds repo top level directory

Yes that had not been done yet, thanks.

alphasentaurii (approved):

PR #1048, opened 2024-07-03
remove breakpoint when test fails

braingram:
@stscieisenhamer found a stray breakpoint in one of the certify test that triggers when the test fails. This PR removes the breakpoint.


braingram:
I would consider this a "no changelog" change but don't have permissions to add labels or request reviewers.

hbrown-ST (approved):

PR #1050, opened 2024-07-05
CCD-1465 Removed all redundant tpn files for roman datamodel-represented references. Certify parameter reference files

stscieisenhamer:
Resolves [CCD-1465](https://jira.stsci.edu/browse/CCD-1465)

There are use cases where parameter reference files will be used with Roman ground processing. Though already implemented for JWST, the certification of such files for Roman is problematic. CRDS had been configured to presume that all references will be represented by Roman datamodels. However, parameter references files are actually more generic, being a feature of the `stpipe` package. This makes it impossible to certify/ingest parameter reference files.

The implementation in this PR is to back out some of the presumption. This involves removing all `tpn` specifications for all references that are Roman datamodel-based. The tpn specs were redundant because datamodel validity is also used by CRDS, making the tpn specifications redundant. This also removes some tech-debt in that the tpn-based validation does not have to be maintained.

This allows more generic tpn specifications, such as those needed for parameter reference files. A new feature has also been implemented. tpn specifications have allowed the phrase "all" to define those tpns that should apply to all reftypes, regardless of instrument. The concept has been expanded to allow any number of "subkinds", delineated by dashes in the reftype name. This is useful for parameter references files, since all parameter reference files are named "<instrument>_pars-<step>". A tpn named "all_pars" now applies to all parameter references, regardless of step or instrument.


alphasentaurii (approved):
Approving this because the test failures appear to be unrelated to any code changes made in this PR and are likely to resolve if the test suite jobs are re-run, perhaps one at a time (in case it has to do with connection/cache issues). This is bolstered by the fact that all tests passed on python 3.11 (mac). I would try re-running the test suite, one job at a time.

PR #1058, opened 2024-08-02
[CI] don't download cache if contexts have not changed

zacharyburnett:
Since the cache is defined by the context, if the combined cache exists with the given name then there is no need to download it, only check its existence with `lookup-only`. If it *doesn't* exist, then save it at the end with the dedicated `actions/cache/save` action.


zacharyburnett:
test run at https://github.com/spacetelescope/crds/actions/runs/10216356086/job/28267657528

hbrown-ST (approved):

PR #1074, opened 2024-09-19
Updated get_build_context to find calver.

hbrown-ST:


alphasentaurii:
Closing this PR - as discussed offline, it is not necessary to include a custom calver optional argument as this should only be determined "under the hood" based on what calver the user has installed locally. Allowing the user to pass in any string here could lead to unintended consequences and reference file incompatibility issues.

PR #1082, opened 2024-10-07
Added several more images and updates for build context.

hbrown-ST:
Resolves [CCD-1508](https://jira.stsci.edu/browse/CCD-1508)

More Doc updates for Data Release work to User Guide

</details>


documentation/crds_users_guide/source/command_line_tools.rst
```diff
@@ -211,7 +211,7 @@ New Context
 
 crds.bestrefs always computes best references with respect to a context which
 can be explicitly specified with the `--new-context` parameter.  If `--new-context`
-is not specified, the default latest context is determined by consulting
+is not specified, the default build context is determined by consulting
```
alphasentaurii, line 214: This is only the case for JWST. This needs to be clarified with something like:  "{...} is not specified, the way the context is determined depends on the observatory.  For Roman and HST, the default latest context is determined by consulting the CRDS server or looking in the local cache. For JWST, the default build context is determined by consulting the CRDS server to find the appropriate context for the calibration software version you have locally installed."

documentation/crds_users_guide/source/library_use.rst
```diff
@@ -181,15 +181,15 @@ crds.get_default_context()
 ..........................
 
 `get_default_context()` returns the name of the pipeline mapping which is 
-currently the Latest Context
+currently the Build Context
```
alphasentaurii, line 184: Same comment as above. `get_default_context` only defaults to build if the observatory is JWST. Otherwise it defaults to latest.

documentation/crds_users_guide/source/library_use.rst
```diff
@@ -181,15 +181,15 @@ crds.get_default_context()
...
 !!!!!!!!!!!!!!!!!!!
 
-*Latest Context* is the .pmap which is nominally in use by the pipeline.
-While it's common to make new files the latest as each context is added, it's
-possible for the *Latest Context* to lag behind the *Edit Context* when
-new files are being added but need additional testing in OPS. Deriving
-from the *Latest Context* is a crude kind of reversion since CRDS
-effectively branches around any existing subsequent contexts. use.
+*Build Context* is the .pmap which is nominally in use by the pipeline.
```
alphasentaurii, line 187: Again, see above. I think it would best to add "build context" info, not replace latest with it. You could add to the first sentence: "*Latest Context* is the .pmap which is nominally in use by the Roman or HST pipeline. For JWST, see "build context" below.

alphasentaurii (changes requested):
Need to clarify that JWST defaults to build context, while Roman and HST default to latest. Both context types are available to all missions, so I don't agree with replacing "latest" with "build" across the board here. We need to include information describing both, and where they apply.

alphasentaurii (approved):
Great - these changes look good to me as far as I can tell. Let's at least start with this and we can make additional adjustments if needed later on.

PR #1086, opened 2024-10-09
add second required argument to jwst get_default_context curl command

braingram:
## Tasks

</details>

### DIFF
--- a/.github/workflows/contexts.yml
+++ b/.github/workflows/contexts.yml
@@ -36,7 +36,7 @@ jobs:
           CRDS_SERVER_URL: https://jwst-crds.stsci.edu
         run: >
           echo "pmap=$(
-          curl -s -X POST -d '{"jsonrpc": "1.0", "method": "get_default_context", "params": ["${{ env.OBSERVATORY }}"], "id": 1}' ${{ env.CRDS_SERVER_URL }}/json/ --retry 8 --connect-timeout 10 |
+          curl -s -X POST -d '{"jsonrpc": "1.0", "method": "get_default_context", "params": ["${{ env.OBSERVATORY }}", null], "id": 1}' ${{ env.CRDS_SERVER_URL }}/json/ --retry 8 --connect-timeout 10 |
           python -c "import sys, json; print(json.load(sys.stdin)['result'])"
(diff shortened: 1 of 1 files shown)


braingram:
The workflow is now working again as seen by a context being found:
https://github.com/spacetelescope/crds/actions/runs/11263290153/job/31320834838?pr=1086#step:5:5
comparing that to the latest run on jwst main:
https://github.com/spacetelescope/jwst/actions/runs/11261422647/job/31314829770#step:5:5

braingram:
I can't request reviewers or add labels. I'd say this is a "no changelog" change and perhaps @alphasentaurii would give this a review.

One open question is if similar server-side API changes are planned for hst and roman. If so, the workflow will need to be updated for those calls (when the servers are updated).

braingram:
I have no idea why the download and cache workflow is failing with:
```
https://github.com/spacetelescope/crds/actions/runs/11263290153/job/31320840346?pr=1086#step:11:4492
```

alphasentaurii:
Sorry just seeing this now - I implemented this exact change for JWST in another PR that was merged last night to resolve related sync errors (see release 12.0.3). Rather than close this PR, if you want you can rebase from master and add the `null` arg to HST and Roman as well. In the mean time, I'll deploy the DRC-style code changes to both of those observatories and then we can merge this in.

alphasentaurii:
> The download and cache workflow is failing: https://github.com/spacetelescope/crds/actions/runs/11263290153/job/31320840346?pr=1086#step:11:4492 Maybe it's related to the change in this PR but I don't at the moment see the link.

When I made this change as part of #1088 the CI didn't succeed until after I merged

zacharyburnett:
should we add `'latest'` to the other two `curl` commands for hst and roman as well?

braingram:
Thanks for taking a look. I'm going to close this PR. jwst and soon romancal and gwcs have modified versions of this workflow. jwst main is currently using `null` as the second argument (whereas it's using `latest` for the workflow in crds main). It might make sense to just ditch the curl commands and use the crds client although a decision about `latest` vs `null` will likely have to be answered for that change as well.

PR #1140, opened 2025-06-12
CCD-

alphasentaurii:
Resolves [CCD-1608](https://jira.stsci.edu/browse/CCD-1608)

If no CRDS environment variables have been set, or if CRDS_SERVER_URL has "serverless" in it (e.g. 'https://jwst-serverless.stsci.edu'), acquire the default latest context from local CRDS cache. It's possible this fallback was not being triggered due to the exception type being raised - additional exceptions have been added to the try-catch condition to subvert an all out error and instead fallback to the local cache gracefully.

## Tasks

</details>

### DIFF
--- a/crds/client/api.py
+++ b/crds/client/api.py
@@ -104,13 +104,19 @@ def set_crds_server(url):
     S = CheckingProxy(URL, version="1.0")
 
 
-def get_crds_server():
+def get_crds_server(obs=None):
     """Return the base URL for the CRDS JSON RPC server.
     """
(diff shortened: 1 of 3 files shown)


crds/client/api.py
alphasentaurii, line 1: removed call to get_server_observatory() because it results in an infinite loop: 
get_default_observatory() --> get_server_observatory() --> get_default_context() --> get_default_observatory()

stscieisenhamer (approved):

PR #1207, opened 2026-05-28
simplify reusable contexts workflow to a GitHub Action for usage in downstream CI

zacharyburnett:
getting the contexts probably should have been a simple action in the first place, instead of a reusable workflow. This change moves the context-grabbing stuff from `.github/workflows/contexts.yml` to `.github/actions/contexts/action.yml` (but keeps the reusable workflow so that downstream usage isn't affected until we can update it to the simpler method)

To migrate from using the reusable workflow to using the action (using `romancal` as an example):
https://github.com/spacetelescope/romancal/blob/26409040b8e777a633ae345281509397bdf22184/.github/workflows/tests.yml#L40-L51
```diff
-  latest_crds_contexts:
-    uses: spacetelescope/crds/.github/workflows/contexts.yml@main
-    with:
-      observatory: roman
   crds_context:
-    needs: [ latest_crds_contexts ]
     runs-on: ubuntu-latest
     steps:
+      - if: inputs.crds_context == ''
+        id: latest_context
+        uses: spacetelescope/crds/.github/actions/contexts@main
+        with:
+          require_roman: true
       - id: context
-        run: echo context=${{ github.event_name == 'workflow_dispatch' && (inputs.crds_context != '' && inputs.crds_context || needs.latest_crds_contexts.outputs.roman) || needs.latest_crds_contexts.outputs.roman }} >> $GITHUB_OUTPUT
+        run: echo context=${{ inputs.crds_context || steps.latest_context.outputs.roman }} >> $GITHUB_OUTPUT
   outputs:
     context: ${{ steps.context.outputs.context }}
```

Current downstream usages that point to `.github/workflows/contexts.yml`:
https://github.com/search?q=org%3Aspacetelescope%20uses%3A%20spacetelescope%2Fcrds%2F.github%2Fworkflows%2Fcontexts.yml&type=code
- `drizzlepac`
- `gwcs`
- `bench`
- `hubble_advanced_products`
- `romancal`
- `romanisim`
- `stenv`
- `roman_datamodels`
- `RegressionTests`

## Tasks

</details>

### DIFF
--- a/.github/actions/contexts/action.yml
+++ b/.github/actions/contexts/action.yml
@@ -1,0 +1,73 @@
+name: crds-latest-contexts
+description: Retrieve latest context filenames for observatories using the Calibration Reference Data System (CRDS)
+branding:
+  icon: book-open
+  color: black
+
+inputs:
+  require_hst:
+    description: require response from Hubble Space Telescope CRDS server
+    required: false
+    default: "false"
+  require_jwst:
+    description: require response from James Webb Space Telescope CRDS server
+    required: false
+    default: "false"
+  require_roman:
+    description: require response from Nancy Grace Roman Space Telescope CRDS server
+    required: false
+    default: "false"
+
+runs:
+  using: composite
+  steps:
+    - id: hst_crds
+      env:
+        OBSERVATORY: hst
+        CRDS_SERVER_URL: https://hst-crds.stsci.edu
+      run: >
+        echo "context=$(
+        curl -s -X POST -d '{"jsonrpc": "1.0", "method": "get_default_context", "params": ["${{ env.OBSERVATORY }}", "latest"], "id": 1}' ${{ env.CRDS_SERVER_URL }}/json/ --retry 8 --connect-timeout 10 |
+        python -c "import sys, json; print(json.load(sys.stdin)['result'])"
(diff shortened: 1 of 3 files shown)


stscieisenhamer:
@zacharyburnett does this require a release to enable in other repos?

zacharyburnett:
> @zacharyburnett does this require a release to enable in other repos?

no it should be usable immediately with 
```yaml
uses: spacetelescope/crds/.github/actions/contexts@main
```

most of the downstream usages use `@main`, but we should probably pin those to the next release for stability at the very least, so it would use a tag in that case 

we could also move `action.yml` to the root of this repo to make it
```yaml
uses: spacetelescope/crds@main 
```

stscieisenhamer (approved):

braingram:
Looks like this broke romancal CI?
https://github.com/spacetelescope/romancal/pull/2219 is blocked since the CI won't start

Can we revert this and make this change in a way that doesn't break downstream usage?

braingram:
Is there an issue, ticket or some other place to discuss these changes?

zacharyburnett:
yes, just made one here: https://jira.stsci.edu/browse/SCSB-262

PR #1198, opened 2026-04-08
Added a check for serverless in the CRDS_SERVER_URL.

hbrown-ST:
Resolves [CCD-nnnn](https://jira.stsci.edu/browse/CCD-1739)

This PR addresses an error encountered in HST when trying to sync while using a cloud path (starting with s3) but configured with CRDS_SERVER_URL set to https://hst-crds-serverless.stsci.edu/.

<details><summary>news fragment change types...</summary>

- ``changes/<PR#>.hst.rst``: HST reference files
</details>

PR #1219, opened 2026-07-12
bugfix/CCD-1766-roman-s3-get

alphasentaurii:
Resolves [CCD-1766](https://jira.stsci.edu/browse/CCD-1766)

- This PR fixes a bug introduced in CRDS client release [13.1.14](https://github.com/spacetelescope/crds/releases/tag/13.1.14) by way of PR #1198 where setting info['connected']=False when CRDS_SERVER_URL contains "serverless" will cause S3 functionality to fail for Roman. Using the "serverless" url directs CRDS not to invoke API calls to the CRDS servers. However, API calls to s3 buckets do require the "connected" attribute to be True.
**FIX**: adds conditional check if observatory == "roman" and the env var S3_ENABLED=1.

- This PR also fixes a minor bug ([CCD-1766](https://jira.stsci.edu/browse/CCD-1766)) relating to s3 mode where the appropriate log messages are not displayed even if downloads/syncs occur. This affects any commands/scripts that rely on `cmdline.ContextsScript.determine_contexts` (i.e. ContextsScript subclasses `crds.sync/SyncScript`, `crds.list/ListScript`)  because the `crds.core.rmap.Mapping._load_mapping` method was preemptively downloading each mapping from s3 during the check to determine which mappings were needed. By the time ContextsScript reached the "official" download step responsible for invoking the `crds_s3_get` plugin, it would see all required mappings are present in the local cache, thereby skipping the entire download sequence and all the expected log messages that go with it. 
**FIX**: modifies `_load_mapping` so that it 1) reads contents of mappings found in the local cache; and 2) if mapping not in cache or `ignore_cache` option = False: performs direct reads of s3 URI contents (no file downloads)

 

## Tasks
- [x] update or add relevant tests
- [x] update relevant docstrings and / or `docs/` page
- [x] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [x] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types)

</details>


hbrown-ST (approved):
LGTM, thanks for adding the S3 tests

